Issue #1, opened 2015-03-11
Create a basic HTTP server


hueniverse:
Let's get started!

First, we need some basic structure: a `lib` folder, a `package.json` file, and an initial `index.js`. Since we will be using **hapi**, we need to include that in the project dependencies.

Second, create a basic **hapi** server on port 8000 which responds to `/version` requests and replies with a simple `{ "version": "0.0.1" }` JSON payload. The version in the response should come from the `package.json` file.

Assignment due: 3/17

Got questions or need help? Post a comment.

AdriVanHoudt:
Maybe a standard way to post PR will be helpful in the future e.g. every PR for this assignment looks like  `[Assignment 1] My PR title`?

rodrigo-medeiros:
@AdriVanHoudt +1

hueniverse:
I'm ok with that. Generally, github links to this issue are enough for me.

rutaihwa:
[ Assignment #1 ] [Basic HTTP server ] (https://github.com/rutaihwa/hueniversity/commit/7d20e56894a13397c7f93747f6c9f385405bc8f9)

hueniverse:
General comments:

- The project structure should follow the same style of every other hapi module. This means `package.json` should include a `main` property that points to either `index.js` or `lib/index.js`. If the project includes a root `index.js`, it must not include any code other than `require('./lib/index.js')`. This makes searching for code easier later as well as keeps the entire application in one spot. I plan to get rid of the root `index.js` file for all the hapi modules I maintain soon as that extra file is useless.
- Most of the submissions had some minors style issues. Pay attention to the empty lines requirement after function declaration as well as not declaring any local variable other than UpperCamelCase required modules and `internals`. See https://gist.github.com/hueniverse/a06f6315ea736ed1b46d for more information.
- While not a worthwhile optimization in this example, there is no reason to have an endpoint that returns a static object response in which that object is constructed every request. The version payload could have been prepared in advance `internals.response = { version: internals.package.version }`.
- If a callback includes an error, check for it! Some of you forgot to check for the error in `server.start(err)`.
- If there is no way to recover from an error and the process should abort, instead of writing custom code to console log it and then exit, just call `Hoek.assert(!err, err)`.
- No need to make up your own `.gitignore`. Just copy it from hapijs/hapi. Also, you only need a root ignore file, not one in every directory.
[...]

PR #2, opened 2015-03-11
Initial HTTP Server

arb:
Closes #1


package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "hueniversity",
+  "version": "1.0.0",
```
calmdev: The assignment calls for a version of `0.0.1` in the payload

index.js
```diff
@@ -0,0 +1,28 @@
+// Load Modules
```
petermilan: I think that this file should contain:
```
module.exports = require('./lib');
```
(the same approach is used inside of hapi)

index.js
```diff
@@ -0,0 +1,28 @@
...
+        handler: function (request, reply) {
+
+            reply({
+                version: internals.package.version
+            });
+        }
+    }
+});
+
+server.start(function () {
```
petermilan: Could we somehow handle the error? The callback should have an err argument:

```

File: index.js [hunk 4]
```diff
@@ -0,0 +1,28 @@
+// Load Modules
+var Hapi = require('hapi');
+
+//Declare internals
+var internals = {
+    package: require('./package.json')
+};
+
+var server = new Hapi.Server();
```
calmdev: Can we use the debug option with `good console`? 

```js
var server = new Hapi.Server({
	debug: {
		request: ['error', 'uncaught']
	}
});
```

index.js
calmdev: What about using `server.log`?

```js
server.log(['error'], 'Failed to start server: ' + error);
server.log(['info'], 'Server running at: ' + server.info.uri);
```

index.js
hueniverse: We're not there yet... no external plugins for now.

index.js
petermilan: looks good for me

index.js
calmdev: Ok

index.js
arb: The problem with using `server.log` is that for this simple example, there isn't anything listening to the events this ultimately emits. Logging to the console is simpler for this first pass.

index.js
```diff
@@ -0,0 +1,33 @@
+// Load Modules
+var Hapi = require('hapi');
+
+//Declare internals
+var internals = {
```
fyockm: Why declare a separate `internals` object? Verses simply:
```
var package: require()
```

AdriVanHoudt: see bottom http://hapijs.com/styleguide

AdriVanHoudt: actual explanation https://gist.github.com/hueniverse/a06f6315ea736ed1b46d#module-classes

index.js
```diff
@@ -0,0 +1,33 @@
...
+
+server.route({
+    method: 'GET',
+    path: '/version',
+    config: {
+        handler: function (request, reply) {
+
+            reply({
+                version: internals.package.version
+            });
```
hueniverse: `return reply()` reads better. Also more efficient to prepare the full response object in advance. In this case it won't make any different but if the static response object was bigger, there is no reason to create and destroy the exact same object over and over again.

index.js
```diff
@@ -0,0 +1,33 @@
...
+        }
+    }
+});
+
+server.start(function (err) {
+
+    if (err) {
+        console.error('There was an error starting the server: ' + err);
+        process.exit(1);
+    }
```
hueniverse: `Hoek.assert(!err)` should be enough and more readable.

lib/index.js
```diff
@@ -0,0 +1 @@
+// Stub file
```
hueniverse: Should be the other way around. The root index.js should never have actual application code. Everything should be in lib (this makes working with IDEs easier when searching for code in the entire project). Either have the package.json file point directly to lib/index.js or use the hapi style of a root index.js that just loads lib/index.js.

package.json
```diff
@@ -0,0 +1,20 @@
...
+  "main": "index.js",
+  "scripts": {
+    "test": "echo 'You should write some tests!'",
+    "start": "node index.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/hueniverse/hueniversity.git"
+  },
+  "license": "ISC",
```
AdriVanHoudt: Shouldn't this be BSD?

PR #5, opened 2015-03-12

TheAlphaNerd:
This is a first pass at implementing a basic HTTP server that has a single route ```/version``` that returns the version from ```package.json```

This PR does not include and automated lintint / testing, but it could easily be added.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,0 +1,29 @@
+var Hapi = require('hapi');
+var Hoek = require('hoek');
+
+var server = new Hapi.Server();
+var internals = {
+  response: {
+    version: require('../package.json').version
+  }
+};
+
+server.connection({ port: 8000 });
+
+server.route({
+  method: 'GET',
+  path: '/version',
+  handler: function (request, reply) {
+    
+    return reply(internals.response);
+  },
+  config: {
+    description: 'Get version of server'
+  }
+});
+
+server.start(function (err) {
+  
+  Hoek.assert(!err, err);
+  console.log('Server running at:', server.info.uri);
+});
--- a/package.json
+++ b/package.json
@@ -1,0 +1,20 @@
+{
+  "name": "hueniversity",
+  "version": "0.0.1",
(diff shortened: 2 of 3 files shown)


TheAlphaNerd:
Closes #1

lib/index.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
```
hueniverse: In general, I don't find `'use strict'` useful. I don't have an objection to it, but we don't include it in hapi modules.

lib/index.js
```diff
@@ -0,0 +1,26 @@
...
+  path: '/version',
+  handler: function (request, reply) {
+    
+    return reply(internals);
+  }
+});
+
+server.start(function (err) {
+  
+  Hoek.assert(!err);
```
hueniverse: Without the second argument, this will print 'Unknown Error'. You want `Hoek.assert(!err, err)`.

package.json
```diff
@@ -0,0 +1,20 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "node lib/index.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/hueniverse/hueniversity/"
+  },
+  "author": "hueniverse",
+  "license": "ISC",
```
hueniverse: hapi uses BSD... :-)

TheAlphaNerd:
@hueniverse All notes fixed

lib/index.js
```diff
@@ -0,0 +1,24 @@
...
+};
+
+server.connection({ port: 8000 });
+
+server.route({
+  method: 'GET',
+  path: '/version',
+  handler: function (request, reply) {
+    
+    return reply(internals);
```
hueniverse: This is dangerous. `internals` is expected to grow as the project evolves. By passing it directly here you are basically using a hack. This will cause surprises later when you expose config and other static info that has nothing to do with it.

TheAlphaNerd: updated to response with internals.response.  I think this might suffer from a similar problem, as there might be multiple responses.  I will be looking through other examples on how to improve this.

PR #7, opened 2015-03-12
[Assignment 1] Basic server

AdriVanHoudt:
Closes #1


index.js
```diff
@@ -0,0 +1,31 @@
+var Hapi = require('hapi');
+
+var internals = {
+	pkg: require('./package.json')
```
toboid: I think submissions should be using the [hapi style guide](https://github.com/hapijs/contrib/blob/master/Style.md), which specifies 4 spaces for rather than tabs for indents

AdriVanHoudt: Thanks for the head up, I'll have to change IDE settings for that, no problem 
And I didn't know that guide existed O_o

toboid: no probs :smile:

index.js
```diff
@@ -0,0 +1,31 @@
...
+
+server.route({
+    method: 'GET',
+    path: '/version',
+    handler: function (request, reply) {
+
+        return reply({ version: internals.pkg.version });
+    },
+    config: {
+        description: 'Returns the version of the server'
```
hueniverse: Nice touch! If you have `config`, it's cleaner to put `handler` inside.

index.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+});
+
+server.start(function (err) {
+
+    if (err) {
+        throw err
```
hueniverse: Ran out of `;`? :-)

index.js
AdriVanHoudt: What do you mean by config? And yes one the routes get more complex the handlers will be moved out of the config.

index.js
AdriVanHoudt: Nice catch :P Is this good for you or do you prefer `Hoek.assert()`?

AdriVanHoudt: I added Hoek, looks cleaner and we're going to need it anyway

index.js
deedubs: I believe he means that it makes sense to do

````
server.route({
    method: 'GET',
    path: '/version',
    config: {
        description: 'Returns the version of the server',
        handler: function (request, reply) {
            return reply({ version: internals.pkg.version });
        }
    }
});
````
I typically move my handlers to a file `./handlers.js`, so it looks like this in my `index.js`

````
server.route({
    method: 'GET',
    path: '/version',
    config: handlers.version
});
````

Comment (AdriVanHoudt) [inline index.js]: Ah I see, yes I also move them to a separate place. Thanks for clearing that up!

Comment (AdriVanHoudt) [inline index.js]: @hueniverse is that a general rule? To put handler in config if it is present? It will probably be cleaner once the handler is in a separate place

Comment (hueniverse) [inline index.js]: That's what I like to do. More readable.

Comment (AdriVanHoudt) [inline index.js]: :+1:

File: index.js [hunk 4]
```diff
@@ -0,0 +1 @@
+module.exports = require('./lib/index.js');
```
hueniverse: No reason to export anything here.

lib/index.js
```diff
@@ -0,0 +1,30 @@
...
+
+server.route({
+    method: 'GET',
+    path: '/version',
+    config: {
+        handler: function (request, reply) {
+
+            return reply({ version: internals.pkg.version });
+        },
+        description: 'Returns the version of the server'
```
hueniverse: `description` better placed before `handler`. Easier to read.

index.js
AdriVanHoudt: just `require('./lib/index.js');` then? Or even `require('./lib')` although i'm not familiar with what it will do then

AdriVanHoudt: Ok found it in the docs will patch

lib/index.js
AdriVanHoudt: lgtm will patch, thanks for all the feedback!

AdriVanHoudt:
@hueniverse awesome thanks! If you want them squashed next time let me know :p

PR #10, opened 2015-03-12
[Assignment #1] Initial server

pruett:
Addresses #1 

- Create Hapi server and add  `/version` route which returns JSON payload

Run `npm .` or `node lib/index.js` to start server

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,285 @@
+{
+  "name": "hueniversity",
+  "version": "0.0.1",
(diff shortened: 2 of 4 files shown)


AdriVanHoudt:
Is a shrinkwrap not overkill at this point?

.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
hueniverse: See hapijs/hapi ignore file for a more complete file to use.

npm-shrinkwrap.json
```diff
@@ -0,0 +1,285 @@
+{
```
hueniverse: There is no need for a shrinkwrap file especially since our only dependency at this point has its own.

PR #14, opened 2015-03-12
[Assignment 1] Basic server

Neablis:


TheAlphaNerd:
Don't think you meant to check in your swap file :smile:

lib/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+server.route({
+    method: 'GET',
+    path:'/version',
+    handler: function (request, reply) {
+       reply({"version": package.version});
+    }
+});
+
+fs.readFile('./package.json', 'utf8', function (err, data) {
```
TheAlphaNerd: this can be replaced with 
```var pkg = require('./package.json');```

hueniverse: And in general, process setup is better off using blocking calls for readability since in practice, you are blocked until we can start the server.

Neablis: @TheAlphaNerd Yes, it can but thought it might be advantage in the future if we build of this project to use FS instead of require. I don't know how yet, just preference. Good catch on my being dumb with swp files, i'm using this course to get better with vim and I forget dumb things like that. 

@hueniverse I'm not in love with using readFileSync in this situation because I would have to add my own try catch block versus relying on the err object. I can see benefits to using the try catch especially if i was doing some global error handler, but this is a lot more intuitive at first look. But yes it does give the impression that the blocking isn't necessary(And if this was production i would probable log the error and return some default version instead of failing altogether)

PR #15, opened 2015-03-13
[Assigment #1] Basic HTTP Server

lvegerano:
initial commit


.nvmrc
```diff
@@ -0,0 +1 @@
+0.10.37
```
hueniverse: Should be in your .gitignore file :-)

lib/index.js
```diff
@@ -0,0 +1,29 @@
...
+    host: 'localhost',
+    port: 8000
+});
+
+server.route({
+    method: 'GET',
+    path: '/version',
+    handler: function (request, reply) {
+
+
```
hueniverse: Just one empty line.

lib/index.js
```diff
@@ -0,0 +1,29 @@
+var pkg = require('../package.json');
+var Hapi = require('hapi');
+
+var server = new Hapi.Server();
+
+server.connection({
+    host: 'localhost',
```
hueniverse: The assignment did not ask to limit the server to `localhost`.

lib/index.js
```diff
@@ -0,0 +1,29 @@
...
+    port: 8000
+});
+
+server.route({
+    method: 'GET',
+    path: '/version',
+    handler: function (request, reply) {
+
+
+        reply({version: pkg.version});
```
hueniverse: You should take a close look at the style guide. Your object is missing some spaces.

PR #19, opened 2015-03-14
Assignment 1

generalhenry:
closes #1

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,11 @@
+var server = module.exports = require('./lib/server');
+
+if (!module.parent) {
+    server.start(function (error) {
+        if (error) {
+            console.error('Error starting server');
+            throw error;
+        }
+        console.log('Server running at:', server.info.uri);
+    });
+}
--- a/lib/server.js
+++ b/lib/server.js
@@ -1,0 +1,6 @@
+var Hapi = require('hapi');
+var server = module.exports = new Hapi.Server();
+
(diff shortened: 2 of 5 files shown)


lib/server.js
```diff
@@ -0,0 +1,6 @@
+var Hapi = require('hapi');
+var server = module.exports = new Hapi.Server();
+
+server.connection({ port: process.env.PORT || 8000 });
+
+server.route(require('./routes'));
```
hueniverse: Better to put all `require()` calls at the top. Makes it easier to know what this file depends on.

lib/server.js
```diff
@@ -0,0 +1,6 @@
+var Hapi = require('hapi');
+var server = module.exports = new Hapi.Server();
```
hueniverse: This should be in the next block since you are not loading dependencies here.

index.js
```diff
@@ -0,0 +1,11 @@
+var server = module.exports = require('./lib/server');
```
hueniverse: Why export the server file which doesn't expose anything?

hueniverse: I prefer `module.exports = internals.server = require('./lib/server');` because it puts the important part first (the interface you are exposing).

index.js
```diff
@@ -0,0 +1,11 @@
...
+
+if (!module.parent) {
```
hueniverse: It's probably cleaner to move execution code to another module that is not used by others to access some interface.

PR #21, opened 2015-03-15
[Assingnment 1] Basic HTTP server

rodrigo-medeiros:


basic-http-server/lib/index.js
```diff
@@ -0,0 +1,11 @@
+var Hapi = require('hapi'),
```
hueniverse: Take a look at how the other pull requests structure the files so when merged it looks like a real module.

basic-http-server/lib/index.js
```diff
@@ -0,0 +1,11 @@
...
+    routes = require('./routes');
```
hueniverse: The style guide requires only one variable assignment per statement.

basic-http-server/lib/routes.js
```diff
@@ -0,0 +1,13 @@
+var version = require('../package.json').version;
+
+var routes = {};
```
hueniverse: Both `version` and `routes` belong inside `internals`.

PR #23, opened 2015-03-15
[Assignment 1 - Create a basic HTTP server]

pltod:


index.js
```diff
@@ -0,0 +1,18 @@
...
+
+server.connection({
+    port: 8000
+});
+
+server.route({
+    method: 'GET',
+    path: '/version',
+    handler: function(request, reply) {
+        reply({ "version": require('./package.json').version });
```
hueniverse: No need to use `"` around simple keys.

PR #24, opened 2015-03-15
[Assignment #1] Create a basic HTTP hapi server

idanwe:
* Add basic hapi server
* Add /version route
* Add .gitignore

Closes #1

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,0 +1,30 @@
+var Hapi = require('hapi');
+var Hoek = require('hoek');
+var Package = require('../package.json');
+
+var internals = {
+    version: Package.version
+};
+
+var server = new Hapi.Server();
+
+server.connection({ port: 8000 });
+
+server.route({
+    method: 'GET',
+    path: '/version',
+    config: {
+        description: 'Returns the server version',
+        handler: function (request, reply) {
+
+            return reply({ version: internals.version });
+        }
+    }
+});
+
+server.start(function (err) {
+
+    Hoek.assert(!err, err);
+
+    console.log('Server is running at ', server.info.uri);
+});
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,15 @@
+.idea
+*.iml
+npm-debug.log
+dump.rdb
+node_modules
+results.tap
+results.xml
+config.json
+.DS_Store
+*/.DS_Store
+*/*/.DS_Store
+._*
+*/._*
+*/*/._*
+coverage.*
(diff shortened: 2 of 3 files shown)


lib/index.js
```diff
@@ -0,0 +1,25 @@
+var Hapi = require('hapi');
+var Package = require('../package.json');
+
+var server = new Hapi.Server();
+server.connection({ port: 3000 });
```
hueniverse: Wrong port.

lib/index.js
```diff
@@ -0,0 +1,28 @@
+var Hapi = require('hapi');
+var Hoek = require('hoek');
+var Package = require('../package.json');
+
+var internals = {
+    version: Package.version
```
idanwe: @hueniverse Is this way of adding properties to the internals follows the hapi coding style?
Compared to:
```javascript
var internals = {};

internals.version = Package.version;
```

AdriVanHoudt: Using `Package` with an uppercase here is confusing since you'd expect it to be a module.
I think it would be cleaner to expose the `package.json` as a whole in `internals` e.g. `internals.package` and using `internals.package.version` in the handler.
More info can be found [here](https://gist.github.com/hueniverse/a06f6315ea736ed1b46d#module-classes)

idanwe: Every file in node behaves as module, and I think the package isn't something internals for index.js.

@hueniverse point out "The package.json file can be assigned to either an UpperCamelCase variable (var Package = require('./package.json');) or to an internals variable but not just a module global per the style guide."

AdriVanHoudt: You're right they do both behave as module. After re reading the styleguide and the gist I'm not fully sure what goes where. 
It says `All require statements must be declared at the top of the module` so by that your code seems good (except I would expose the whole package but that's just minor).
It also states `Every module can only have two top level globals (exports and internals)`. I don't know what is meant with globals since `var hapi = require('hapi')` would also be a module global not?

As seen in https://github.com/hapijs/hapi/blob/master/lib/plugin.js it looks like you're right.
Now I ask if doing `internals.package = require('../package.json);` is wrong to do? @hueniverse

AdriVanHoudt: I talked to @marsup and he cleared some things up for me. 
* The styleguide is not clear. What is meant that the only module globals can be requires (like hapi or package), internals and exports.
* This means that what you're doing is just fine (sorry for the big responses :angel: )
* To go back to your original question, either is fine ;)

hueniverse: Either one is fine. And feel free to PR improvements to the style guide.

AdriVanHoudt: @hueniverse I did a PR can you take a look? https://github.com/hapijs/contrib/pull/23

PR #25, opened 2015-03-16
Create basic server

ivanph:
- Adds basic server 

Entry for assignment #1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,15 @@
+.idea
+*.iml
+npm-debug.log
+dump.rdb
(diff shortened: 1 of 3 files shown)


lib/version.js
```diff
@@ -0,0 +1,3 @@
+var Package = require('../package.json');
+
+module.exports.number = Package.version;
```
hueniverse: This extra file doesn't add any value since all it does is load a file and exposes part of it. Better of to just require the package.json file directly.

PR #26, opened 2015-03-16
feat(server): version route

MrBri:
Resolve #1 
testing might be premature
editor settings via editorconfig
eslint settings via package.json to match lab
watch server file changes via pm2

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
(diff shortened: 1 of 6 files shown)


MrBri:
Just saw recent comment from @hueniverse. Code requires update.

.editorconfig
```diff
@@ -0,0 +1,11 @@
+# EditorConfig is awesome: http://EditorConfig.org
```
hueniverse: This file should be in your .gitignore

test/server.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
```
hueniverse: Nice! But let's hold off adding tests just yet.

PR #27, opened 2015-03-16
assigment 1  : basic http server 

roelof1967:


roelof1967:
I apolize for this. I thought it would be merged into my earlier push.

hueniverse:
@roelof1967 pull requests are free.

lib/index.js
```diff
@@ -0,0 +1,21 @@
...
+var server = new Hapi.Server(); 
+
+server.connection({
+   port: 8000  
+ }); 
+  
+ server.route({
+     method: 'GET', 
+     path: '/version', 
+     handler: function (request, reply) {
```
hueniverse: Missing new line.

lib/index.js
```diff
@@ -0,0 +1,21 @@
...
+         reply({ version: file.version })
```
hueniverse: better to do `return reply(...)`

lib/index.js
```diff
@@ -0,0 +1,21 @@
...
+  
+ server.route({
+     method: 'GET', 
+     path: '/version', 
+     handler: function (request, reply) {
+         reply({ version: file.version }) 
+    } 
+ }); 
+  
+server.start(function () {
```
hueniverse: Missing new line. Missing `err` in callback, and checking for errors.

lib/index.js
roelof1967: Oke, can you give me a tip where I have missed it. I do not see it.

lib/index.js
roelof1967: oke, where did I miss a new line. The callback function I will implement.

lib/index.js
idanwe: "Newline after { except for inlined or empty objects"
https://github.com/hapijs/contrib/blob/master/Style.md#newlines

PR #29, opened 2015-03-16
Create a basic server

oluoluoxenfree:
hueniverse/hueniversity#1

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,7 +1,9 @@
 var Hapi = require('hapi');
-var internals = require('../package.json');
+var Hoek = require('hoek');
 
-internals.response = { version: internals.version };
+var internals = {};
+internals.Json = require('../package.json');
+internals.response = { version: internals.Json.version };
 
 var server = new Hapi.Server();
 server.connection({ port: 8000 });
@@ -10,11 +12,11 @@ server.route({
     method: 'GET',
     path: '/version',
     handler: function (request, reply) {
-        reply(internals.response)
-          .type('application/json');
+        return reply(internals.response).type('application/json');
     }
 });
 
-server.start(function () {
+server.start(function (err) {
+    Hoek.assert(!err, err);
     console.log('Server running at:', server.info.uri);
 });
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -0,0 +1,20 @@
+var Hapi = require('hapi');
+var internals = require('../package.json');
```
idanwe: internals should defined as object separately from the modules

lib/index.js
```diff
@@ -0,0 +1,20 @@
...
+internals.response = { version: internals.version };
+
+var server = new Hapi.Server();
+server.connection({ port: 8000 });
+
+server.route({
+    method: 'GET',
+    path: '/version',
+    handler: function (request, reply) {
+        reply(internals.response)
```
idanwe: add return

lib/index.js
```diff
@@ -0,0 +1,20 @@
...
+          .type('application/json');
+    }
+});
+
+server.start(function () {
```
idanwe: Add err handling

lib/index.js
oluoluoxenfree: How do you mean? 

The internals thing has confused me a bit but googling it hasn't brought up much.

idanwe: Look at [eran's node.js coding style on the modules section] (https://gist.github.com/hueniverse/a06f6315ea736ed1b46d#module-classes)

lib/index.js
```diff
@@ -0,0 +1,22 @@
+var Hapi = require('hapi');
+var Hoek = require('hoek');
+
+var internals = {};
+internals.Json = require('../package.json');
```
hueniverse: If you are going to keep the raw package file, don't use Uppercase variable name here (e.g. use `internals.json` instead).

lib/index.js
```diff
@@ -0,0 +1,22 @@
...
+internals.response = { version: internals.Json.version };
+
+var server = new Hapi.Server();
+server.connection({ port: 8000 });
+
+server.route({
+    method: 'GET',
+    path: '/version',
+    handler: function (request, reply) {
+        return reply(internals.response).type('application/json');
```
hueniverse: Missing empty line after function declaration.

PR #32, opened 2015-03-16

adrian-tiberius:


lib/index.js
```diff
@@ -0,0 +1,40 @@
+var Hapi = require('hapi');
+var Hoek = require('hoek');
+
+var internals = {
+    Package: require('../package.json'),
```
AdriVanHoudt: use uppercase only for imported modules outside of internals

lib/index.js
```diff
@@ -0,0 +1,39 @@
...
+var internals = {
+    port: 8000,
+    responses: {
+        version: {
+            version: Package.version
+        }
+    }
+};
+
+internals.Server = function () {
```
hueniverse: What's the value of making this method a constructor?

adrian-tiberius:
I was thinking it will be easier to move the server functionality to a different file later, if I have all the server functionality grouped into that function.

PR #33, opened 2015-03-16

Couto:
Closes #1

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,28 +1,37 @@
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var internals = {
-    version: require('../package.json').version
+var internals = {};
+
+internals.package = require('../package.json');
+
+internals.response = {
+    version: internals.package.version
 };
 
-var server = new Hapi.Server();
+internals.Server = function () {
 
-server.connection({
-    port: 8000
-});
+    this.server = new Hapi.Server();
+    this.server.connection();
+};
 
-server.route({
+exports.server = internals.server = new internals.Server();
+
+internals.server.route({
     path: '/version',
     method: 'GET',
     config: {
+        description: 'Output server\'s version.',
         handler: function (request, reply) {
 
-            return reply(internals.version);
+            return reply(internals.response);
         }
     }
 });
 
-server.start(function (err) {
+exports.server.connection({ port: 8000 });
+
+internals.server.start(function (err) {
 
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -0,0 +1,37 @@
+var Hapi = require('hapi');
+var Hoek = require('hoek');
+var internals = {};
```
AdriVanHoudt: a new line between modules and module global vars makes it easier to read

adrian-tiberius:
I think there is an issue with the code. exports.server and internals.server, which reference the same object, doesn't have a route or connection function. the object that does have those functions is exports.server.server and internals.server.server. One solution which I think would also make the code more readable would be to move the route and connection inside the Server class you defined on line 11 which contains the hapi server object ( this.server ).

lib/index.js
```diff
@@ -0,0 +1,37 @@
...
+
+internals.package = require('../package.json');
+
+internals.response = {
+    version: internals.package.version
+};
+
+internals.Server = function () {
```
AdriVanHoudt: Using a capital here is not allowed see http://hapijs.com/styleguide
Try moving `server` out of internals this will make your code way simpler

Couto: I'm still battling with the styleguide. :)
Where do you suggest that I place the `server`? Should i keep it in the `exports` only? (I'm ok with it, just asking)

Couto:
@adrian-tiberius You're absolutely right! I'll fix that as soon as I get home! :) 
Thanks for the tip, really not sure how I didn't notice that! :+1:

lib/index.js
AdriVanHoudt: I am to struggling with the style guide(mostly the use of internals :p) 
I'm waiting for @hueniverse's response to clarify things up :D

lib/index.js
```diff
@@ -0,0 +1,37 @@
...
+    config: {
+        description: 'Output server\'s version.',
+        handler: function (request, reply) {
+
+            return reply(internals.response);
+        }
+    }
+});
+
+exports.server.connection({ port: 8000 });
```
hueniverse: Calling `connection()` after `route()` means the route will not be added to the connection. Did you try running this?

Couto: I did, but then somehow I screwed up! I have no idea what I was doing last night.
I've just pushed another commit fixing this.

PR #34, opened 2015-03-17
Assignment 1 solution

Mkoopajr:


lib/routes.js
```diff
@@ -0,0 +1,18 @@
...
+};
+
+module.exports = [
+    {
+        method: 'GET',
+        path: '/version',
+        config: {
+            handler: function (request, reply) {
+
+                return reply({version: internals.version});
```
AdriVanHoudt: missing some spaces for a 1 lin object

hueniverse:
This is nicely done. One comment is that I generally recommend to avoid "forward looking" code organizations. By that I mean, you have one route and already created a routes.js file. It is trivial to move routes to such a file at a later point, but right now, it just makes your code more complex than it should be.

PR #36, opened 2015-03-17
Assignment #1

MrBri:
fix #1


index.js
```diff
@@ -0,0 +1 @@
+module.exports = require('./lib');
```
AdriVanHoudt: no reason to export here, made the same "mistake" ;)

package.json
```diff
@@ -0,0 +1,115 @@
+{
+  "name": "hueniversity",
+  "version": "0.0.2",
```
AdriVanHoudt: should be 0.0.1

package.json
```diff
@@ -0,0 +1,115 @@
...
+  "scripts": {
+    "start": "node .",
+    "watch": "pm2 start index.js --watch --name hueniversity",
+    "stop": "pm2 stop --watch hueniversity",
+    "test": "lab -t 100 -L",
+    "test-cov-html": "lab -r html -o coverage.html"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/MrBri/hueniversity.git"
```
AdriVanHoudt: should point to upstream (hueniverse)

AdriVanHoudt:
Tests were not a part of the assignment, great habit though!
Also no need for watch, stop and eslint configs in `package.json`

lib/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+server.route([{
+    method: 'GET',
+    path: '/version',
+    config: {
+        description: 'version from package.json',
+        handler: function(request, reply) {
+
+            return reply({ version: internals.pack.version })
+                .type('application/json');
```
hueniverse: No need to specify `type` manually. hapi will automatically set it.

package.json
```diff
@@ -0,0 +1,115 @@
...
+  "dependencies": {
+    "hapi": "^8.4.0",
+    "hoek": "^2.11.1"
+  },
+  "devDependencies": {
+    "code": "^1.3.0",
+    "lab": "^5.5.0",
+    "pm2": "^0.12.7"
+  },
+  "eslintConfig": {
```
hueniverse: If you really want to specify lint rules, use an rc file instead of putting it in package.json. Cleaner.

test/server.js
```diff
@@ -0,0 +1,27 @@
...
+var it = lab.it;
+var before = lab.before;
+var after = lab.after;
+var expect = Code.expect;
+
+describe('server', function() {
+
+    it('should return a version', function(done) {
+
+        var server = require('../lib/index');
```
hueniverse: This is fine for now but will become a problem since the test has little control over the creation of the server.

AdriVanHoudt: wouldn't that be covered by Lab?

hueniverse: Well, it will be handled when we restructure this app in the next assignment...

PR #38, opened 2015-03-17
assignment 1

chyld:


AdriVanHoudt:
* Test were not a part of the assignment (it is good habit though!) also when using lab try to use hapi/code for `expect`.
* When you have a config in the route it is cleaner to put the handler in the config to
* not fully sure but server shouldn't be in `internals` pinging @hueniverse for that

adrian-tiberius:
I think server should be under internals because, unless I misunderstood, the style info states that only the internals global variable is allowed (https://gist.github.com/hueniverse/a06f6315ea736ed1b46d under Module Classes).

AdriVanHoudt:
@adrian-tiberius true the style guide does say that. But I have yet to see a Hapi example (on the website or any Hapi repo) that does it. I did a PR on the style guide to add examples to clarify some things https://github.com/hapijs/contrib/pull/23, feel free to add to it or the start a conversation about it. 
@hueniverse can you clarify here or on the PR? I feel like a lot of people are confused about the internals object

PR #40, opened 2015-03-17
[Assigment 1] Create http server

osukaa:
Closes #1

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,0 +1,30 @@
+var Hapi = require('hapi');
+var Hoek = require('hoek');
+
+var package = require('../package.json');
+
+var internals = {
+    staticVersion: {
+        version: package.version
+    }
+};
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -0,0 +1,30 @@
+var Hapi = require('hapi');
+var Hoek = require('hoek');
+
+var package = require('../package.json');
```
AdriVanHoudt: The package is basically a module so the name should start with a capital letter `var Package = ...`

lib/index.js
```diff
@@ -0,0 +1,30 @@
...
+
+var internals = {
+    staticVersion: {
```
hueniverse: `staticVersion` isn't a very helpful name. `versionResponse`, `response`, etc. are more descriptive.

PR #42, opened 2015-03-18
Homework Assignment #1

towns:
I know I am late, still looking for feedback. One question I have is I have 4 space tabs for index.js, but what about package.json?


AdriVanHoudt:
I commented on a commit since you merged with master and the result is no changes :P

hueniverse:
JSON files can have 4 or 2 indentation. Typically 2 but I've seen 4. The style guide does not care about it since it's not code, just a data file.

Issue #43, opened 2015-03-18
Convert to a plugin


hueniverse:
The right way to work with hapi is to build your application using plugins. Plugins provide a way to organize your code into logical components and then put them together in different combinations or deploy them in different configurations. While some plugins are published as general purpose utilities (e.g. adding authentication), you should think of plugins as a way to break your code into pieces.

Now that we have a basic server with one endpoint, we want to move the creation of the endpoint to a [plugin](http://hapijs.com/tutorials/plugins), and then register that plugin with the server. Creating a new file `lib/version.js` and move the `/version` endpoint there, wrapped in the plugin `register()` function.

Then change our current `lib/index.js` to require the new file and register the version plugin with our server.

Remember to follow the style guide, and ask any questions right here in the comments. If you are not sure how to get your fork back in sync with the current updated code, use the [git guide](https://github.com/hueniverse/hueniversity/blob/master/guides/git.md).

Assignment due: 3/24

AdriVanHoudt:
Nice git guide!

AdriVanHoudt:
How far do we need to go with the plugin? e.g. add its own package.json and tests?

TheAlphaNerd:
I was under the impression that since we were asked to make ```lib/version.js``` that would be the extent of what we are to do for a plugin.  @hueniverse had previously mentioned testing would be coming in a future assignment (although I do long for tests!).

I was also noticing the amount of noise that my pr has created referencing this issue... is this a problem?

AdriVanHoudt:
yeah I guess and I noticed the noise to. How come? Looks like the same commit every time?

TheAlphaNerd:
OHHHHHH I think it is from my rebasing

AdriVanHoudt:
aaah probably yes

TheAlphaNerd:
Btw

```plus = commit -a --amend --no-edit```

FTW!!!

AdriVanHoudt:
for adding to an existing commit?

TheAlphaNerd:
mmhmmm.  That alias will take all staged chages and squash them down into the last commit.

hueniverse:
No need for a standalone plugin module, just the one file.

AdriVanHoudt:
:+1:

roelof1967:
I have now this.

index.js : 

~~~~

// Load modules

var Hapi = require('hapi');
var Hoek = require('hoek');
var Package = require('../package.json');


// Declare internals

var internals = {};


[...]

PR #44, opened 2015-03-18
[assignment 2] Convert version route to plugin

TheAlphaNerd:
Closes #43

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,35 +2,24 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
-
+var Version = require('./version');
 
 // Declare internals
 
 var internals = {};
 
-
 internals.init = function () {
 
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
-
-    server.route({
-        method: 'GET',
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
-
-                return reply({ version: Package.version });
-            }
-        }
-    });
-
-    server.start(function (err) {
-
+    server.register(Version, function (err) {
+      
         Hoek.assert(!err, err);
-        console.log('Server started at: ' + server.info.uri);
+        server.start(function (err) {
+
+            Hoek.assert(!err, err);
+            console.log('Server started at: ' + server.info.uri);
+        });
     });
 };
 
--- a/lib/version.js
+++ b/lib/version.js
@@ -1,0 +1,29 @@
+// Declare internals
+var internals = {
+  response: {
+    version: require('../package.json').version
+  }
+};
+
+// Register plugin
+exports.register = function (server, options, next) {
+
+    server.route({
+        method: 'GET',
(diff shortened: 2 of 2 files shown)


lib/version.js
```diff
@@ -0,0 +1,27 @@
+// Declare internals
+var internals = {
+  response: {
+    version: require('../package.json').version
+  }
+};
+
+// Register plugin
+exports.register = function (server, options, next) {
+    server.route({
```

lib/version.js
```diff
@@ -0,0 +1,27 @@
...
+        method: 'GET',
+        path: '/version',
+        config: {
+            description: 'Returns the version of the server',
+            handler: function (request, reply) {
+
+                return reply(internals.response);
+            }
+        }
+    });
```
AdriVanHoudt: I would put a newline here for readability

TheAlphaNerd: done did it

lib/version.js
TheAlphaNerd: got it!

lib/version.js
AdriVanHoudt: :+1:

lib/version.js
AdriVanHoudt: :+1:

lib/version.js
```diff
@@ -0,0 +1,29 @@
...
+
+    next();
```
hueniverse: This is a callback so `return`

TheAlphaNerd: Good call!

sschuck:
Do we want to check if the version is a valid semver?

AdriVanHoudt:
@sschuck the version comes from `package.json` and that should always follow semver

lib/index.js
```diff
@@ -2,31 +2,22 @@
...
 var Hoek = require('hoek');
-var Package = require('../package.json');
-
 
 // Declare internals
 
-var internals = {};
-
+var internals = {
+  version: require('./version')
```
hueniverse: This is odd. Just `var Version = require('./version');` at the top and then use that in `register()`.

TheAlphaNerd: updated

lib/index.js
```diff
@@ -2,35 +2,25 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
```
hueniverse: No empty line here.

TheAlphaNerd: In personal projects I have always put a new line between modules that are included from ```node_modules``` and modules that are being included locally from the project.

That aside the style has been fixed.

AdriVanHoudt: I also like to do this, makes it easier to see what are local modules

hueniverse: I just sort them: node, npm, local.

TheAlphaNerd: Reasonable... I follow the same pattern (with spaces).

Thanks for taking the tie to discuss

AdriVanHoudt: same sorting here only with newlines in between

PR #45, opened 2015-03-18
assignment 2

chyld:


chyld:
Closes issue #43

lib/version.js
```diff
@@ -0,0 +1,29 @@
...
+            }
+        }
+    });
+
+    return next();
+};
+
+exports.register.attributes = {
+    name: 'version',
+    version: '0.0.1'
```
hueniverse: No need to specify versions when it is a built in plugin inside your app.

lib/index.js
```diff
@@ -2,35 +2,23 @@
...
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
-
-                return reply({ version: Package.version });
-            }
-        }
-    });
-
-    server.start(function (err) {
+    server.register(require('./version'), function (err) {
```
hueniverse: Move the `require()` to the top like any other loaded module.

PR #46, opened 2015-03-18
Assignment #2: restructured project to use plugin for version endpoint

towns:


lib/index.js
```diff
@@ -2,28 +2,20 @@
...
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
+    server.register(require('./version'), function (err) {
 
-                return reply({ version: Package.version });
-            }
+        if (err) {
+            console.error("Plugin didn't load properly:", err);
```
AdriVanHoudt: I would also use `Hoek.assert` here like in `server.start`

lib/index.js
```diff
@@ -2,29 +2,19 @@
...
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
 
-    server.route({
-        method: 'GET',
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
+    server.register(require('./version'), function (err) {
```
hueniverse: Move `require()` to the top like any other loaded module.

PR #47, opened 2015-03-18
Move route to plugin  assigment #2

edwin-bluekite:


lib/index.js
```diff
@@ -2,29 +2,21 @@
...
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
 
 
 // Declare internals
 
 var internals = {};
-
+internals.version = require('./version');
```
hueniverse: Assign it to `Version`.

lib/index.js
```diff
@@ -2,29 +2,21 @@
...
-
-                return reply({ version: Package.version });
-            }
-        }
+    server.register(internals.version, function (err) {
+	
+	Hoek.assert(!err, err);
     });
 
     server.start(function (err) {
```
hueniverse: `start()` needs to be inside the `register()` callback.

lib/version.js
```diff
@@ -0,0 +1,22 @@
...
+            }
+        }
+    });
+
+    return next();
+}
+
+exports.register.attributes = {
+      name: 'version'
+}
```
hueniverse: Missing `;`

edwin-bluekite:
Updated with all suggestions

lib/index.js
```diff
@@ -2,35 +2,24 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
 
 
-// Declare internals
-
+var Version = require('./version');
```
hueniverse: This should be with the modules section above.

PR #49, opened 2015-03-19
[Assignment #2] Convert to a plugin 

Couto:
Closes #43

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -23,7 +23,6 @@ internals.init = function () {
             console.log('Server started at: ' + server.info.uri);
         });
     });
-
 };
 
 internals.init();
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,8 +2,6 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
-
 
 // Declare internals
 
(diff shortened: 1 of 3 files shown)


lib/version.js
```diff
@@ -0,0 +1,25 @@
...
+            description: 'Returns the version of the server',
+            handler: function (request, reply) {
+
+                return reply({ version: Package.version });
+            }
+        }
+    });
+
+    return next();
+
```
AdriVanHoudt: no newline required here

lib/version.js
```diff
@@ -0,0 +1,24 @@
...
+        }
+    });
+
+    return next();
+};
+
+
+exports.register.attributes = {
+    name: 'version',
+    version: Package.version
```
hueniverse: No need for `version` when the plugin is internal to the app.

lib/index.js
```diff
@@ -15,22 +13,15 @@ internals.init = function () {
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
 
-    server.route({
-        method: 'GET',
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
+    server.register({ register: require('./version.js') }, function (err) {
```
hueniverse: No need to wrap the plugin in `{register}`. Also move the `require()` up like any other module.

lib/version.js
Couto: That makes perfect sense, specially because I'm using the package's version.

Not directly related to this code, but is it usual to find routes with independent versioning?
example: `/version` is currently 0.0.2, but `/users` could be 0.0.5? Just asking out of curiosity.

hueniverse: It's common but that means moving those plugins to their own modules, not internal like we are doing here.

PR #50, opened 2015-03-19
[Assignment 2] Convert to a plugin

AdriVanHoudt:
Closes #43


lib/version.js
```diff
@@ -0,0 +1,26 @@
...
+            }
+        }
+    });
+
+    return next();
+};
+
+exports.register.attributes = {
+    name: 'version',
+    version: '0.1.0'
```
hueniverse: Where did you get this version number from?

AdriVanHoudt: top of my head :) and also semver, I'll patch, with the comments you made on the issue in mind, later

AdriVanHoudt:
I'll squash the commits if you want to merge

lib/index.js
```diff
@@ -15,22 +15,15 @@ internals.init = function () {
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
 
-    server.route({
-        method: 'GET',
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
+    return server.register(Version, function (err) {
```
chyld: why are you using return? i checked and most submissions are not using it.

lib/index.js
```diff
@@ -15,22 +15,15 @@ internals.init = function () {
...
 
-                return reply({ version: Package.version });
-            }
-        }
-    });
+        Hoek.assert(!err, err);
 
-    server.start(function (err) {
+        return server.start(function (err) {
```
chyld: return also

lib/index.js
AdriVanHoudt: not super needed here but it makes it clear where the functions returns, i make it a habit to prevent derpy errors like calling functions after it should have returned etc

lib/index.js
AdriVanHoudt: same as above

PR #51, opened 2015-03-19
Converting to plugin

rutaihwa:
[ Assignment 2 - Converting to plugin ] #43

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,7 +2,7 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var versionPlugin = require('./version');
+var Version = require('./version');
 
 // Declare internals
 
@@ -14,11 +14,7 @@ internals.init = function () {
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
 
-    server.register({register: versionPlugin }, {
-	routes: {
-            prefix: '/version'
-	}
-    }, function (err) {
+    server.register({register: Version }, {} , function (err) {
 
         Hoek.assert(!err, err);
     });
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,7 +2,6 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
 var versionPlugin = require('./version');
 
 // Declare internals
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,7 +3,7 @@
 var Hapi = require('hapi');
 var Hoek = require('hoek');
 var Package = require('../package.json');
-
+var versionPlugin = require('./version');
 
 // Declare internals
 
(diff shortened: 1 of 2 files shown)


lib/version.js
```diff
@@ -0,0 +1,25 @@
...
+        config: {
+	    description: 'Returns the version of the server',
+	    handler: function (request, reply) {
+		
+                return reply({ version: Package.version });
+	    }
+        }
+    });
+    
+    next();
```
AdriVanHoudt: `next` is a callback so better to do `return next()`

lib/index.js
```diff
@@ -3,7 +3,7 @@
 var Hapi = require('hapi');
 var Hoek = require('hoek');
 var Package = require('../package.json');
```
AdriVanHoudt: you are not using package in index anymore so no need to require it

rutaihwa: Yeah, i overlooked that. Thanks @AdriVanHoudt

AdriVanHoudt: np :+1:

lib/index.js
```diff
@@ -2,8 +2,7 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
-
+var versionPlugin = require('./version');
```
hueniverse: `Version`

lib/index.js
```diff
@@ -15,16 +14,13 @@ internals.init = function () {
...
 
-    server.route({
-        method: 'GET',
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
+    server.register({register: versionPlugin }, {
+	routes: {
+            prefix: '/version'
```
hueniverse: Why use the prefix here instead of just setting the route path directly?

rutaihwa: Incase, we'd have different versions like v1, v2 etc?

lib/index.js
```diff
@@ -15,16 +14,9 @@ internals.init = function () {
...
-            handler: function (request, reply) {
+    server.register({register: Version }, {} , function (err) {
 
-                return reply({ version: Package.version });
-            }
-        }
+        Hoek.assert(!err, err);
     });
 
     server.start(function (err) {
```
edwin-bluekite: Same as my error, you should start the server in the register callback

lib/version.js
```diff
@@ -0,0 +1,25 @@
...
+	    description: 'Returns the version of the server',
+	    handler: function (request, reply) {
+		
+                return reply({ version: Package.version });
+	    }
+        }
+    });
+    
+    return next();
+}
```
edwin-bluekite: Add semicolon

lib/version.js
```diff
@@ -0,0 +1,25 @@
...
+
+exports.register.attributes = {
+    name: 'versionPlugin',
+    version: '0.0.1'
```
edwin-bluekite: Version is not needed here

lib/index.js
hueniverse: This is a lot of complexity for no gain. Don't optimize for unknown future needs.

lib/version.js
rutaihwa: @edwin-bluekite where am I missing the semicolon again?

lib/index.js
rutaihwa: You are right! :+1:  Thanks.

PR #52, opened 2015-03-19
Convert to a plugin

gmertk:


lib/version.js
```diff
@@ -0,0 +1,29 @@
...
+		}
+	});
+
+	return next();
+};
+
+
+exports.register.attributes = {
+	name: 'version',
+	version: '0.0.1'
```
hueniverse: No need for version here.

PR #53, opened 2015-03-19
[Assignment2] Convert to a plugin.

zoe-1:
Credit to:
Myles Borins
TheAlphaNerd
https://github.com/TheAlphaNerd
This commit is modeled after his assignment 2 code below:
https://github.com/TheAlphaNerd/hueniversity/commit/c0d9d04723922632ac56be213609d9ee11e8bd6a

I learned a lot from his code.
The distributed classroom is working :-)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,7 +3,6 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
 
 // Declare internals
 var internals = {
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,30 +1,25 @@
+// ./lib/index.js 
 // Load modules
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
 var Package = require('../package.json');
 
-
 // Declare internals
-
-var internals = {};
+var internals = {
+    version: require('./version')
+};
 
 
 internals.init = function () {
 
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -1,30 +1,25 @@
+// ./lib/index.js 
 // Load modules
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
 var Package = require('../package.json');
```
AdriVanHoudt: Not used anymore here

lib/index.js
```diff
@@ -1,30 +1,25 @@
+// ./lib/index.js
```
AdriVanHoudt: Why is this here?

zoe-1: Personal habit.  When working on project in vim I think it is helpful to have this note.

lib/index.js
zoe-1: Thanks. I removed it.

lib/index.js
AdriVanHoudt: But since this is a public repo and not everyone uses vim this is unnecessary

zoe-1: Good point.  I removed the comment.

AdriVanHoudt: :+1:

hueniverse:

PR #54, opened 2015-03-19
[Assignment #2] Convert to a plugin

pruett:
Reference issue #43

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,36 +2,26 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
-
 
 // Declare internals
 
 var internals = {};
 
-
 internals.init = function () {
 
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
-
-    server.route({
-        method: 'GET',
-        path: '/version',
-        config: {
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -1,30 +1,19 @@
...
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
-
-                return reply({ version: Package.version });
-            }
-        }
+        Hoek.assert(!err, err);
     });
 
     server.start(function (err) {
```
AdriVanHoudt: you are starting your server without being sure the plugin is fully registered, take a look at it

lib/version.js
```diff
@@ -0,0 +1,18 @@
+var Package = require('../package.json');
+
+exports.register = function (server, options, next) {
+    server.route({
```

lib/version.js
```diff
@@ -0,0 +1,20 @@
...
+
+                return reply({ version: Package.version });
+            }
+        }
+    });
+
+    return next();
+};
+
+exports.register.attributes = { name: 'versionRoute' };
```
hueniverse: Cleaner to put object on multiple lines.

hueniverse:

PR #56, opened 2015-03-19
[Assignment #2] Refactor out version to plugin

idanwe:
Closes #43

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -1,0 +1,33 @@
+// Load modules
+
+var Package = require('../package.json');
+
+// Declare internals
+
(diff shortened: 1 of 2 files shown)


lib/version.js
```diff
@@ -0,0 +1,26 @@
...
+            }
+        }
+    });
+
+    return next()
+}
+
+exports.register.attributes = {
+    name: 'version',
+    version: Package.version
```
AdriVanHoudt: The version here is the version of the plugin so not the version of the server

hueniverse:
Perfect.

PR #58, opened 2015-03-20
Refactor version route using a plugin

eatrero:
Looking forward to learning and hapi-ness. Thanks Hueniversity!

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -15,15 +15,14 @@ internals.init = function () {
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
 
-    server.route({
-        method: 'GET',
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
-
-                return reply({ version: Package.version });
-            }
+    server.register([
+        {
+            register: require('./version'),
+            options: Package
+        }
+    ], function (err) {
+        if (err) {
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -15,15 +15,14 @@ internals.init = function () {
...
-            handler: function (request, reply) {
-
-                return reply({ version: Package.version });
-            }
+    server.register([
+        {
+            register: require('./version'),
+            options: Package
+        }
+    ], function (err) {
```
AdriVanHoudt: newline after function, also try using the same error handling as in `server.start()`

AdriVanHoudt: Also you are starting the server whether the registration of the plugin errors or not, might want to look into that

lib/version.js
```diff
@@ -0,0 +1,16 @@
...
+            handler: function (request, reply) {
+                return reply({ version: options.version });
+            }
+        }
+    });
+};
+
+exports.register.attributes = {
+    pkg: { name : 'version plugin' }
+};
```
AdriVanHoudt: newline at the end of a file

lib/version.js
```diff
@@ -0,0 +1,16 @@
+exports.register = function (server, options, next) {
+    server.route({
```
AdriVanHoudt: newline after function also in the handler

PR #59, opened 2015-03-21
Assignment #2 [Convert to plugin]

lvegerano:
Converted version route to a plugin

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -14,7 +14,7 @@ exports.register = function (server, options, next) {
         }
     });
 
-    next();
+    return next();
 };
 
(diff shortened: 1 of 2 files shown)


lib/version.js
```diff
@@ -0,0 +1,23 @@
...
+        config: {
+            description: 'Returns the version of the server from a plugin',
+            handler: function (request, reply) {
+
+                return reply({ version: Package.version });
+            }
+        }
+    });
+
+    next();
```
hueniverse: `return next();`

hueniverse:

PR #61, opened 2015-03-22
Move route version into its own plugin.

osukaa:
Closes #43

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,12 +2,12 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
-
 
 // Declare internals
 
-var internals = {};
+var internals = {
+    version: require('./version')
+};
 
 
 internals.init = function () {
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -15,16 +15,9 @@ internals.init = function () {
...
-            handler: function (request, reply) {
+    server.register(internals.version, function (err) {
 
-                return reply({ version: Package.version });
-            }
-        }
+        Hoek.assert(!err,err);
     });
 
     server.start(function (err) {
```
chyld: you should start your server _after_ the plugins have been registered. it needs to be placed in the server.register callback.

lib/index.js
```diff
@@ -2,12 +2,12 @@
...
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
-
 
 // Declare internals
 
-var internals = {};
+var internals = {
+    version: require('./version')
```
idanwe: It is a module you should assign it to `Version` and then register it

PR #63, opened 2015-03-22
Assignment 2 - extract version route to plugin

toboid:
#43


lib/version.js
```diff
@@ -0,0 +1,33 @@
...
+
+
+// Declare internals
+
+var internals = {};
+
+
+internals.versionInfo = { version: Package.version };
+
+module.exports.register = function (server, options, next) {
```
chyld: you can just say exports.register, no need to add module

lib/version.js
```diff
@@ -0,0 +1,33 @@
...
+                return reply(internals.versionInfo);
+            }
+        }
+    });
+
+    return next();
+};
+
+
+module.exports.register.attributes = {
```
chyld: remove module, just exports.register

lib/index.js
```diff
@@ -15,22 +15,14 @@ internals.init = function () {
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
 
-    server.route({
-        method: 'GET',
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
+    server.register(Version, function (registerErr) {
```
hueniverse: Unless you need to access errors on multiple levels, just call it `err`. It's a standard.

lib/version.js
toboid: thanks

lib/index.js
toboid: will do, thanks

PR #64, opened 2015-03-22
Convert to a plugin

gitting:
The two first commits should be squashed prior to pulling.
Well, all of them, actually.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -15,9 +15,7 @@ internals.init = function () {
     server.connection({ port: 8000 });
 
     server.register({ register: require('./version') }, function (err) {
-        if (err) {
-            console.error('Failed to load plugin:', err);
-        }
+        Hoek.assert(!err, err);
     });
 
     server.start(function (err) {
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -15,16 +14,8 @@ internals.init = function () {
...
-        method: 'GET',
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
-
-                return reply({ version: Package.version });
-            }
-        }
+    server.register({ register: require('./version') }, function (err) {
```
chyld: move module loading to top of file

lib/version.js
```diff
@@ -0,0 +1,21 @@
...
+                return reply({ version: Package.version });
+            }
+        }
+    });
+
+    next();
+};
+
+exports.register.attributes = {
+    pkg: Package
```
hueniverse: Internal plugins should just have `name`.

hueniverse:

gitting:
Is there a naming convention for internal plugins? Should `multiple: true` ever be used for internal plugins `attributes`?

hueniverse:
I usually use the file name as the plugin name unless there is conflict. As for `multiple`, that's based on the nature of the plugin.

PR #65, opened 2015-03-22
Assignment 2: Convert to a plugin

ipoddubny:


lib/version.js
```diff
@@ -0,0 +1,30 @@
+// Load modules
+
+var Package = require('../package.json');
+
+
+// Declare internals
+
+var internals = {};
```
edwin-bluekite: internals is declared but not used

TheAlphaNerd: prior art --> https://github.com/hueniverse/hueniversity/blob/master/lib/index.js#L10

hueniverse: `internals` should always be declared. Saves you from forgetting later.

lib/version.js
```diff
@@ -0,0 +1,30 @@
...
+                return reply({ version: Package.version });
+            }
+        }
+    });
+
+    return next();
+};
+
+exports.register.attributes = {
+    name: '/version route handler'
```
hueniverse: `name` should be access friendly (e.g. no spaces, or characters you need to put in quotes).

PR #67, opened 2015-03-23
assigment2 attempt 1

sfabriece:


lib/index.js
```diff
@@ -2,9 +2,9 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
+var Version = require('./version');
 var Package = require('../package.json');
```
chyld: you're not using the Package module in this file

lib/index.js
```diff
@@ -15,16 +15,11 @@ internals.init = function () {
...
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
-
-                return reply({ version: Package.version });
-            }
-        }
+    server.register(Version, function(err){
+       if(err){
```
chyld: use Hoek for error checking

lib/version.js
```diff
@@ -0,0 +1,24 @@
+//Load modules
+var Package = require('../package.json');
+
+//Declare internals
+var internals = {};
```
chyld: you've declared internals, but are not using it

hueniverse: That's the correct style. Always declare `internals`.

AdriVanHoudt: Even when not used? Seems redundant. Also it will throw a linting error for unused vars...

lib/index.js
sfabriece: The reply is getting the Version number from Package.

PR #68, opened 2015-03-23
Assignment 2: Convert to a plugin

rodrigo-medeiros:
Could've used the 'load one plugin' method to register, but since probably we're going to build more plugins, I found relevant to use the other one (passing an array).

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,8 +2,7 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
-
+var Version = require('./version');
 
 // Declare internals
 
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -15,15 +14,15 @@ internals.init = function () {
...
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
+    server.register([
+        {
+            register: Version,
+            options: {}
+        }
+    ], function (err) {
```
chyld: use Hoek

hueniverse:
Don't try to predict the future. You made you code more complex now, when adding `[]` later will take literally seconds.

lib/index.js
rodrigo-medeiros: Good catch @chyld, fixed, thanks.

rodrigo-medeiros:
I see @hueniverse, you're right. But, to be honest, I think that passing an array of objects to the `register` method makes the code much more _readable_, even though I know this is totally subjective.

PR #71, opened 2015-03-23
Assignment2

mikejerome:
Attempting assignment 2

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -21,10 +21,10 @@ internals.init = function () {
 
         Hoek.assert(!err, err);
 
-        return server.start(function (err) {
+        server.start(function (err) {
 
             Hoek.assert(!err, err);
-            return console.log('Server started at: ' + server.info.uri);
+            console.log('Server started at: ' + server.info.uri);
         });
     });
 };
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,7 +2,7 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
+var Version = require('./version');
 
 
 // Declare internals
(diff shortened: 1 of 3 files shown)


mikejerome:
Hopefully I've followed the style guide properly. I'm quite new and the style guide was a bit daunting for me. I'm mainly wondering about my blank lines and return statements.

lib/index.js
```diff
@@ -15,22 +15,17 @@ internals.init = function () {
...
-        }
-    });
-
-    server.start(function (err) {
+    server.register(Version, function (err) {
 
         Hoek.assert(!err, err);
-        console.log('Server started at: ' + server.info.uri);
+
+        return server.start(function (err) {
```
chyld: no need for return

lib/index.js
```diff
@@ -15,22 +15,17 @@ internals.init = function () {
...
+
+            Hoek.assert(!err, err);
+            return console.log('Server started at: ' + server.info.uri);
```
chyld: console.log returns undefined

lib/version.js
```diff
@@ -0,0 +1,24 @@
...
+exports.register = function (server, options, next) {
+
+    server.route({
+        method: 'GET',
+        path: '/version',
+        config: {
+            description: 'Returns the version of the server',
+            handler: function (request, reply) {
+
+                return reply({ version: Package.version });
```
chyld: you may want to cache the response in an internals variable

mikejerome:
Thanks @chyld.

lib/version.js
gyaresu: Sorry to jump in here but is it just a matter of declaring `var internals = Package.version;` and if so why?

gyaresu:
Sorry to jump in here but I read the comment about moving to an 'internals' variable. Is it just a matter of declaring var internals = { version: Package.version }; and if so why? 

Also, is it redundant to `require` package.json again in the version.js file? (As I've also done)

AdriVanHoudt:
@gyaresu there is no real reason to store the version in internals since you can directly access it through Package. 
You still have to require package.json in version.js so you can reply with the server version

lib/version.js
```diff
@@ -0,0 +1,33 @@
...
+exports.register = function (server, options, next) {
+
+    server.route({
+        method: 'GET',
+        path: '/version',
+        config: {
+            description: 'Returns the version of the server',
+            handler: function (request, reply) {
+
+                return reply(internals.version);
```
toboid: The previous version returned a payload in the format `{ version: 0.0.1 }`, you have changed the behaviour here to return just a version number with no enclosing object (this will just return `0.0.2`)

gyaresu: Are you sure? Running it live returns what I expect: {"version":"0.0.2"}

mikejerome:
@chyld Could you please explain what you mean by caching the response in an internals variable?

lib/version.js
mikejerome: Fixed, thanks.

PR #72, opened 2015-03-24
[Assignment #2] Refactor out version to plugin

cwilgenhoff:
Requesting CR.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,35 +2,31 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
+var Version = require('./version.js');
 
 
 // Declare internals
 
-var internals = {};
+var internals = {
+    connection: {
+        port: process.env.PORT || 8000
+    }
+};
 
 
 internals.init = function () {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection(internals.connection);
 
-    server.route({
-        method: 'GET',
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
-
-                return reply({ version: Package.version });
-            }
-        }
-    });
-
-    server.start(function (err) {
+    server.register(Version, function (err) {
 
         Hoek.assert(!err, err);
-        console.log('Server started at: ' + server.info.uri);
+        server.start(function (err) {
+    
+            Hoek.assert(!err, err);
+            console.log('Server started at: ' + server.info.uri);
+        });
     });
 };
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,7 +9,7 @@ var Version = require('./version.js');
 
 var internals = {
     connection: {
-        port: process.env.PORT || 8000
+        port: 8000
     }
 };
 
@@ -29,5 +29,5 @@ internals.init = function () {
         });
     });
 };
-
+s
 internals.init();
--- a/lib/version.js
+++ b/lib/version.js
@@ -1,0 +1,33 @@
+// Load modules
+
+var Package = require('../package.json');
+
+// Declare internals
+
+var internals = {
+    response: {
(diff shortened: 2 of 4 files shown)


lib/index.js
```diff
@@ -2,35 +2,31 @@
...
-                return reply({ version: Package.version });
-            }
-        }
-    });
-
-    server.start(function (err) {
+    server.connection(internals.connection);
 
+    server.register(Version, function (err) {
         Hoek.assert(!err, err);
```
chyld: newline after function

lib/index.js
```diff
@@ -2,35 +2,31 @@
...
-        console.log('Server started at: ' + server.info.uri);
+    
+        return server.start(function (err) {
```
chyld: why are you using return?

cwilgenhoff: My bad.

lib/version.js
```diff
@@ -0,0 +1,33 @@
...
+        config: {
+            description: 'Returns the version of the server',
+            handler: function (request, reply) {
+
+                return reply(internals.response);
+            }
+        }
+    });
+    
+    next();
```
AdriVanHoudt: use `return next()` for readability

lib/version.js
```diff
@@ -0,0 +1,33 @@
...
+};
+
+exports.register.attributes = {
+    name: 'version'
+};
```
AdriVanHoudt: newline at the end of the file

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "hueniversity",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "description": "Community learning experiment",
-  "main": "lib/index.js",
```
AdriVanHoudt: why change this?

lib/version.js
TheAlphaNerd: fwiw I believe the return serves a pragmatic purpose beyond readability.  It also ensures that the context / stack frame of this function is appropriately exited when the callback is executed.  This ensures that no weird behavior happens such as unexpected data manipulation, race conditions, or even a second callback call (EEP)

AdriVanHoudt: you are right it also prevents calling callbacks twice and stuff, although in this particular case that wouldn't happen but I completely agree, I use it where I can, in this case even in the init function in index.js

package.json
cwilgenhoff: This is due I've added an index file into the base project folder, so no need to go inside lib to run it.

AdriVanHoudt: that behaviour was specifically removed from solution of the first assignment and is not included in the second one so no need to change it

lib/version.js
cwilgenhoff: I don't understand the need of a return, as you said, in this particular case callbacks are not getting called more than once at a time.

package.json
cwilgenhoff: Oh, sorry my bad for not having started since assignment 1. Thanks for the heads up.

AdriVanHoudt: Np, try to only modify what is in the assignment and not add extra stuff (like this or e.g. tests or something) when not asked ;)

lib/version.js
AdriVanHoudt: In this case readability, when quickly reading the code I can immediately see where the callback is called take a look at https://gist.github.com/hueniverse/a06f6315ea736ed1b46d#return-callbacks

lib/index.js
```diff
@@ -2,35 +2,31 @@
...
-var Package = require('../package.json');
+var Version = require('./version.js');
 
 
 // Declare internals
 
-var internals = {};
+var internals = {
+    connection: {
+        port: process.env.PORT || 8000
```
hueniverse: Adding env support wasn't part of the assignment.

cwilgenhoff: Sorry, I was developing on c9, that was the reason to use it. Removed.

PR #73, opened 2015-03-24
Second assignment, convert route /version to a plugin

omarvides:
Closes #43

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -7,9 +7,7 @@ var Version = require('./version');
 
 // Declare internals
 
-var internals = {
-    version : Version,
-};
+var internals = {};
 
 
 internals.init = function () {
@@ -17,7 +15,7 @@ internals.init = function () {
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
 
-    server.register(internals.version, function (err) {
+    server.register(Version, function (err) {
         
         Hoek.assert(!err, 'Failed to load plugin', err);
     });
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,12 +2,14 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var version = require('./version');
+var Version = require('./version');
 
 
 // Declare internals
 
-var internals = {};
+var internals = {
+    version : Version,
+};
 
 
 internals.init = function () {
@@ -15,11 +17,9 @@ internals.init = function () {
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
 
-    server.register({register: version}, function (err) {
-
-        if (err) {
-            console.error('Failed to load plugin:', err);
-        }
+    server.register(internals.version, function (err) {
+        
+        Hoek.assert(!err, 'Failed to load plugin', err);
     });
 
     server.start(function (err) {
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,7 +2,7 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
+var version = require('./version');
 
 
 // Declare internals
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -2,7 +2,7 @@
 
 var Hapi = require('hapi');
 var Hoek = require('hoek');
-var Package = require('../package.json');
+var version = require('./version');
```
mikejerome: Capitalize Version according to style guide

lib/index.js
```diff
@@ -15,15 +15,10 @@ internals.init = function () {
...
-        method: 'GET',
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
+    server.register({register: version}, function (err) {
 
-                return reply({ version: Package.version });
-            }
+        if (err) {
```
mikejerome: use hoek

lib/version.js
```diff
@@ -0,0 +1,23 @@
...
+            }
+        }
+    });
+
+    next();
+};
+
+exports.register.attributes =  {
+    name: 'versionPlugin',
+    version: '1.0.0'
```
mikejerome: You don't need version here because it's for internal use

lib/version.js
```diff
@@ -0,0 +1,23 @@
...
+exports.register = function (server, option, next) {
+    
+    server.route({
+        method: 'GET',
+        path: '/version',
+        config: {
+            description: 'Returns the version of the server',
+            handler: function (request, reply) {
+
+                return reply({ version: Package.version });
```
mikejerome: consider moving Package.version to internals

lib/index.js
```diff
@@ -15,15 +15,10 @@ internals.init = function () {
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
 
-    server.route({
-        method: 'GET',
-        path: '/version',
-        config: {
-            description: 'Returns the version of the server',
-            handler: function (request, reply) {
+    server.register({register: version}, function (err) {
```
mikejerome: you can pass Version directly instead of {register: version}

lib/version.js
AdriVanHoudt: also you can just use the filename as name for the plugin

lib/version.js
AdriVanHoudt: @mikejerome why?

lib/version.js
```diff
@@ -0,0 +1,23 @@
...
+            }
+        }
+    });
+
+    next();
```
AdriVanHoudt: use `return reply()` here for readability

lib/version.js
mikejerome: I may be wrong here I was just passing on feedback from my own submission but I think I may have misunderstood. Is it possible to cache the response here? It was suggested to possibly cache the response in internals.

AdriVanHoudt: @mikejerome I don't think storing your response will help here since it is an object with only 1 key. Besides it makes is harder to read since the response of your handler is stored outside of it.

mikejerome: @AdriVanHoudt Ok thanks. I changed it back in my code as well. Do you have an example of what that would look like just out of curiosity?

AdriVanHoudt: @mikejerome how what would look like?

mikejerome: @AdriVanHoudt Caching the respose

AdriVanHoudt: @mikejerome I guess not having to remake to response object counts as caching. Since you would only define the object once and pass it to the handler multiple times you don't have to rebuild it but in this case it's neglectable

omarvides:
Thanks for the feedback! :)

lib/index.js
```diff
@@ -2,12 +2,14 @@
...
 var Hoek = require('hoek');
-var Package = require('../package.json');
+var Version = require('./version');
 
 
 // Declare internals
 
-var internals = {};
+var internals = {
+    version : Version,
```
hueniverse: This is not needed.

PR #74, opened 2015-03-24
refactor(version): use of plugin system

MrBri:
fix #43

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -14,7 +14,7 @@ exports.register = function(server, options, next) {
         }
     });
 
-    next();
+    return next();
 };
 
 exports.register.attributes = {
--- a/lib/version.js
+++ b/lib/version.js
@@ -1,0 +1,22 @@
+var Package = require('../package.json');
+
+exports.register = function(server, options, next) {
+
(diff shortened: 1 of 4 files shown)


lib/version.js
```diff
@@ -0,0 +1,22 @@
...
+        config: {
+            description: 'Returns the version of the server',
+            handler: function (request, reply) {
+
+                return reply({ version: Package.version });
+            }
+        }
+    });
+
+    next();
```
AdriVanHoudt: use `return next()` here for readability

package.json
```diff
@@ -1,22 +1,23 @@
 {
   "name": "hueniversity",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "description": "Community learning experiment",
   "main": "lib/index.js",
   "repository": {
     "type": "git",
-    "url": "https://github.com/hueniverse/hueniversity.git"
+    "url": "https://github.com/MrBri/hueniversity.git"
```
AdriVanHoudt: the urls in here should always point to upstream repo(hueniverse's)

PR #75, opened 2015-03-24
Assignment 2 - first attempt

gyaresu:


lib/index.js
```diff
@@ -3,7 +3,7 @@
 var Hapi = require('hapi');
 var Hoek = require('hoek');
 var Package = require('../package.json');
```
AdriVanHoudt: Package is not used anymore

lib/version.js
```diff
@@ -0,0 +1,20 @@
+var Package = require('../package.json');
+
+exports.register = function (server, options, next) {
+    server.route({
```

lib/version.js
```diff
@@ -0,0 +1,20 @@
...
+        method: 'GET',
+        path: '/version',
+        config: {
+            description: 'Returns the version of the server',
+        handler: function (request, reply) {
+            return reply({ version: Package.version});
```
AdriVanHoudt: newline after function and some indentations

Issue #79, opened 2015-03-24
100% coverage


hueniverse:
Things are getting a bit more interesting...

It's time to add tests, verify coverage, confirm style, and automate all of this with CI. We will be using the [lab](https://github.com/hapijs/lab) module to perform all these tasks and automate it with [travis](https://travis-ci.org).

1. Add a `.travis.yml` file, testing our project on node 0.10, 0.12, and io.js (latest).
2. Add a test folder with two files, `version.js` and `index.js`, each testing the corresponding file under `/lib`.
3. Modify the `package.json` file to include the tests, as well as the dev dependency to lab.
4. Add the standard hapi.js Makefile to make it easy to generate other types of reports (e.g. HTML coverage report).
5. When using lab, enable coverage, require 100% coverage, enable linting with default rules, and use the [code](https://github.com/hapijs/code) assertion library.
6. Write a basic test to verify our version endpoint in `version.js`.
7. Change the `init()` method to accept a port and a callback. Use the callback to return when the function completes or errors. The `init()` callback should return any error state as well as a reference to the newly created server. This will allow us to later stop the server when we test it.
8. Export `init()` and move the invocation to a new `start.js` file (which will call the `init()` function with the `8000` port and a callback the outputs the information to the console when started). Change the `package.json` file to use the `start.js` file as the starting place. This file will not be covered by tests.
9. Write a basic test to verify the `init()` function in `index.js`.
10. Bring coverage to 100% by adding more tests as needed.

Everything up to (10) should be pretty straight forward. If you are not sure on how to use lab and code, look at other hapi.js modules like hoek, qs, items, and boom (e.g. simple modules) to copy their test scripts and setup.

Getting 100% coverage can be tricky sometimes so if you are not sure, get as much coverage as you can, and comment on the lines in your pull request where you are having a hard time reaching and someone will give you a clue.

Remember to properly `stop()` your servers when calling the `init()` method in each test.

For now, avoid using any of the `before()` and `after()` lab features.

As always, ask for help and help others!

Due: 4/4

AdriVanHoudt:
Wasn't there a general agreement(if you want I'll search for the issue in question) to use npm for all the different tests as in define them in package.json instead of a Makefile?

hueniverse:
1. Fixed.
2. The makefile gives you easy access to stuff like `make cov-html` while the package.json one is for `npm test`.

AdriVanHoudt:
1. Ok but you still declare the response in internals, why?
[...]

PR #80, opened 2015-03-25
assignment 3; code coverage

chyld:
PR for issue #79 

At 93% code coverage. I'm using NODE_ENV because the `init()` method in `index.js` was getting called twice. Once when the file was being required, and once in the test. This was causing port contention issues when no port was specified.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,10 +10,12 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+// Initialization
+
+exports.init = internals.init = function (port, cb) {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port || 8000 });
     server.register(Version, function (err) {
 
         Hoek.assert(!err, err);
@@ -21,8 +23,14 @@ internals.init = function () {
 
             Hoek.assert(!err, err);
             console.log('Server started at: ' + server.info.uri);
+
+            if (cb) {
+                return cb(err, server);
+            }
         });
     });
 };
 
-internals.init();
+if (process.env.NODE_ENV !== 'test') {
+    internals.init();
+}
--- a/lib/index.js
+++ b/lib/index.js
@@ -13,11 +13,11 @@ exports.init = internals.init = function (port, cb) {
 
     if (typeof port === 'function') {
         cb = port;
-        port = 8000;
+        port = null;
     }
 
     var server = new Hapi.Server();
-    server.connection({ port: port });
+    server.connection({ port: port || 8000 });
     server.register(Version, function (err) {
 
         if (err) {
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,36 +1,31 @@
 // Load modules
 
 var Hapi = require('hapi');
-var Hoek = require('hoek');
 var Version = require('./version');
-
 
 // Declare internals
 
 var internals = {};
 
-
 // Initialization
 
 exports.init = internals.init = function (port, cb) {
 
+    if (typeof port === 'function') {
+        cb = port;
+        port = 8000;
+    }
+
     var server = new Hapi.Server();
-    server.connection({ port: port || 8000 });
+    server.connection({ port: port });
     server.register(Version, function (err) {
 
-        Hoek.assert(!err, err);
+        if (err) {
+            return cb(err, server);
+        }
+
         server.start(function (err) {
-
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
-
-            if (cb) {
-                return cb(err, server);
-            }
+            return cb(err, server);
         });
     });
 };
-
-if (process.env.NODE_ENV !== 'test') {
-    internals.init();
-}
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,13 +3,9 @@
 var Hapi = require('hapi');
 var Version = require('./version');
 
-// Declare internals
-
-var internals = {};
-
 // Initialization
 
-exports.init = internals.init = function (port, cb) {
+exports.init = function (port, cb) {
 
     if (typeof port === 'function') {
         cb = port;
--- a/lib/index.js
+++ b/lib/index.js
@@ -5,23 +5,19 @@ var Version = require('./version');
 
 // Initialization
 
-exports.init = function (port, cb) {
-
-    if (typeof port === 'function') {
-        cb = port;
-        port = null;
-    }
+exports.init = function (port, next) {
 
     var server = new Hapi.Server();
-    server.connection({ port: port || 8000 });
+    server.connection({ port: port});
     server.register(Version, function (err) {
 
         if (err) {
-            return cb(err, server);
+            return next(err, server);
         }
 
         server.start(function (err) {
-            return cb(err, server);
+
+            return next(err, server);
         });
     });
 };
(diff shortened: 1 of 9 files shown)


lib/index.js
```diff
@@ -10,10 +10,12 @@ var Version = require('./version');
...
+// Initialization
+
+exports.init = internals.init = function (port, cb) {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port || 8000 });
     server.register(Version, function (err) {
 
         Hoek.assert(!err, err);
```
TheAlphaNerd: I think we are supposed to cb this error rather than assert if a callback is given

lib/index.js
```diff
@@ -21,8 +23,14 @@ internals.init = function () {
...
+
+            if (cb) {
+                return cb(err, server);
+            }
         });
     });
 };
 
-internals.init();
+if (process.env.NODE_ENV !== 'test') {
```
TheAlphaNerd: I'm not sure if this is needed at all... it is my assumption that now with the test suite we are expecting lib to export a server ready to go.

chyld: I think @hueniverse wants us to keep the init method, but export it so it can be called in the test.

lib/index.js
```diff
@@ -21,8 +23,14 @@ internals.init = function () {
 
             Hoek.assert(!err, err);
```
TheAlphaNerd: if you always assert here you will never get to the callback

test/version.js
```diff
@@ -0,0 +1,34 @@
...
+describe('version', function () {
+
+    it('gets the version', function (done) {
+
+        Server.init(4000, function(err, server) {
+
+            server.inject('/version', function(response) {
+
+                expect(response.statusCode).to.equal(200);
+                expect(response.result.version).to.equal('0.0.3');
```
TheAlphaNerd: you should probably reference the package.json here or this test is doomed to fail every semver

chyld:
I'm at about 96% code coverage. I need the plugin registration to fail - but exporting the version plugin from index.js seems like a hack just to increase code coverage.

lib/index.js
```diff
@@ -1,28 +1,31 @@
...
 
+// Initialization
+
+exports.init = internals.init = function (port, cb) {
 
-internals.init = function () {
+    if (typeof port === 'function') {
+        cb = port;
+        port = 8000;
+    }
```
TheAlphaNerd: you don't account for the case of port being null

lib/index.js
```diff
@@ -1,28 +1,31 @@
...
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port });
```
TheAlphaNerd: this will blow up if port is null

chyld: good catch

zoe-1:
@chyld Possibly, you already know what I am about to write below.
But, after doing more research and dialoguing with @TheAlphaNerd I have learned:
That using a a callback to handle errors like your script does allows you 
to write "throw safe" code passing errors up the call back chain.

@TheAlphaNerd says:
If you are writing a plugin or module that is going to be consumed by someone else 
you don't know what they want to do with that error.  Very likely they don't want 
that error to make their entire application asplode.
I tend to try and write "throw safe" code, passing errors up the callback chain. 
Now in theory you should be able to do
"Try / catch" but that isn't terribly async.

zoe-1:
@chyld Thank you for your comments on my assignment3.
They helped a lot. I developed a solution for getting 100% coverage without exporting
stuff from lib/index.js just to pass tests. I think this solution will work for your too. 
1.) In your test/index.js file require your version plugin. 
var Version = require('../lib/version');
2.) Then write tests for coverage of your init() method in test/index.js.
@TheAlphaNerd has written good tests to provide coverage. However, he relies
on data exported just to get tests to pass.

To fix this just change his (@TheAlphaNerd):
lib.Version.register to Version.register as I did after adding the 
Version = require('../lib/version'); described above .

Then, you can get 100% test coverage without exporting data just to get your tests to pass.

Please let me know what you think.

lib/index.js
```diff
@@ -1,28 +1,27 @@
 // Load modules
 
 var Hapi = require('hapi');
-var Hoek = require('hoek');
 var Version = require('./version');
 
+// Initialization
 
-// Declare internals
+exports.init = function (port, cb) {
```
hueniverse: The hapi style requires `callback` or `next` based on the next tick guarantee of the API. In this case it is a `next` since `register()` uses `next` and you didn't wrap the error handling in `Hoek.nextTick()`.

lib/index.js
```diff
@@ -1,28 +1,27 @@
...
 
-var internals = {};
-
-
-internals.init = function () {
+    if (typeof port === 'function') {
+        cb = port;
+        port = null;
+    }
```
hueniverse: The optional argument behavior was removed from the assignment after the initial posting.

lib/index.js
```diff
@@ -1,28 +1,27 @@
...
-        Hoek.assert(!err, err);
-        server.start(function (err) {
+        if (err) {
+            return cb(err, server);
+        }
 
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+        server.start(function (err) {
+            return cb(err, server);
```
hueniverse: New line.

package.json
```diff
@@ -22,7 +22,12 @@
     "hoek": "2.x.x"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "start": "node lib/index.js"
+    "test": "lab -La code -t 100",
+    "test-cov-html": "lab -La code -r html -o coverage.html",
+    "start": "node start.js"
```
hueniverse: `start.js` should live in `lib`

start.js
```diff
@@ -0,0 +1,12 @@
+// Load modules
+
+var Hoek = require('hoek');
+var Server = require('./lib');
```
hueniverse: Not sure `Server` is the best variable name here.

test/version.js
```diff
@@ -0,0 +1,29 @@
...
+var lab = exports.lab = Lab.script();
+var describe = lab.experiment;
+var it = lab.test;
+var expect = Code.expect;
+
+describe('version', function () {
+
+    it('gets the version', function (done) {
+
+        Server.init(4000, function(err, server) {
```
hueniverse: You should avoid specifying ports in tests to avoid errors due to conflicts. Instead, use port `0`.

start.js
AdriVanHoudt: When doing `Server.init` it makes sense, doing `Lib.init` or `Index.init` it makes it harder to read imo

test/index.js
```diff
@@ -0,0 +1,43 @@
...
+            server.stop(done);
+        });
+    });
+
+    it('should fail plugin registration', function (done) {
+
+        var register = Version.register;
+
+        Version.register = function(server, options, next) {
+            next('Plugin Error');
```
hueniverse: new line. `return next(...)`

test/index.js
```diff
@@ -0,0 +1,43 @@
...
+    it('should create a server', function (done) {
+
+        Lib.init(0, function(err, server) {
+
+            expect(server.info.port).to.be.above(0);
+            server.stop(done);
+        });
+    });
+
+    it('should fail plugin registration', function (done) {
```
hueniverse: When tests monkey patch code, they should be labeled with `{ parallel: false }` option on `it()`.

PR #81, opened 2015-03-25
Assignment 3

### LINKED ISSUE #79: 100% coverage

TheAlphaNerd:
Closes #79

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,7 +1,6 @@
 // Load modules
 
 var Hapi = require('hapi');
-var Hoek = require('hoek');
 var Version = require('./version');
 
 
@@ -10,19 +9,19 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = function (port, next) {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port });
     server.register(Version, function (err) {
 
-        Hoek.assert(!err, err);
+        if (err) {
+            return next(err, server);
+        }
+
         server.start(function (err) {
 
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+            return next(err, server);
         });
     });
 };
-
-internals.init();
--- a/lib/start.js
+++ b/lib/start.js
@@ -1,8 +1,8 @@
 var Hoek = require('hoek');
 
-var lib = require('./lib');
+var Lib = require('./lib');
 
-lib.init(function (err, server) {
+Lib.init(function (err, server) {
 
     Hoek.assert(!err, err);
     console.log('Server started at: ' + server.info.uri);
--- a/lib/start.js
+++ b/lib/start.js
@@ -1,0 +1,9 @@
+var Hoek = require('hoek');
+
+var lib = require('./lib');
+
+lib.init(function (err, server) {
+
+    Hoek.assert(!err, err);
+    console.log('Server started at: ' + server.info.uri);
+});
--- a/package.json
+++ b/package.json
@@ -22,7 +22,12 @@
     "hoek": "2.x.x"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "start": "node lib/index.js"
+    "test": "lab -La code -t 100",
+    "test-cov-html": "lab -La code -r html -o coverage.html",
+    "start": "node lib/start.js"
+  },
+  "devDependencies": {
+    "code": "^1.3.0",
+    "lab": "^5.5.1"
   }
 }
--- a/test/index.js
+++ b/test/index.js
@@ -1,0 +1,57 @@
+// Load modules
+
+var Code = require('code');
+var Lab = require('lab');
+var lib = require('../lib');
+var Version = require('../lib/version');
+
+
+// Declare internals
+
+var internals = {
+    port: 1337
+};
+
+
+// Test shortcuts
+
+var lab = exports.lab = Lab.script();
+var describe = lab.describe;
+var it = lab.it;
+var expect = Code.expect;
+
+describe('lib', function() {
+
+    describe('.init()', function () {
+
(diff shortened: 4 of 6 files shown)


zoe-1:
From the assignment instructions:
4.) Add the standard hapi.js Makefile to make it easy to generate other types of reports (e.g. HTML coverage report).
You forgot to add the Makefile.

TheAlphaNerd:
There was a debate if the make file is necessary.  The package.json is literally doing the exact same thing.  There is prior art in other hapi projects where make has been replaced with a package.json as make is not supported on windows.

I'll wait for @hueniverse to chime in.

Thanks for the heads up though!!!

hueniverse:
I like having a Makefile. Not a big deal either way.

package.json
```diff
@@ -22,7 +22,12 @@
     "hoek": "2.x.x"
```
chyld: Bump your version to 0.0.3

package.json
```diff
@@ -22,7 +22,12 @@
...
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "lab -La code -t 100",
+    "test-cov-html": "lab -La code -r html -o coverage.html",
     "start": "node lib/index.js"
```
chyld: main: and start: should point to start.js

TheAlphaNerd: oops I thought I updated that... good eye!

package.json
TheAlphaNerd: I've been letting @hueniverse bump the versions.  Generally I think it makes sense for the version bump to happen when it is locked, not during a PR.

lib/index.js
```diff
@@ -10,19 +9,26 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.Version = Version;
```
chyld: I see you exported Version because you wanted to make the plugin registration fail in one of your tests and therefore get 100% code coverage. It has been my practice that you should never modify "production" code just to make a test pass. My question: is there a way to make the plugin registration fail without exporting? @hueniverse Any hints?

package.json
TheAlphaNerd: oh wait... main should still be lib/index, as that is what exports our module.  I did need to update the start script though!

lib/index.js
TheAlphaNerd: it was hinted on IRC that the production code would need to be carefully manipulated in order to get the tests to pass.  I'm not 100% that this was the best way to do it, looking forward to the insight.

chyld: Yes, I totally understand why you did it. But it doesn't feel right to me, i.e., exporting something your wouldn't necessarily export, just to increase code coverage. Hope @hueniverse chimes in.

TheAlphaNerd: Well in theory I could imagine someone requiring this module only wanting access to the plugin� so it isn�t TOTALLY insane.

:P

package.json
chyld: I made main and start both point to start.js. Not sure they should be different.

AdriVanHoudt:
@hueniverse can we do both?

package.json
AdriVanHoudt: good point! :+1:

package.json
TheAlphaNerd: there is no point in having main point to start.  It doesn't export anything

chyld: good point

lib/index.js
idanwe: I think it is better to require the `'../lib/version'` at the test instead of exports `Version` here

chyld: I got 100% coverage without having to export version from the index.js. Node caches the first require. So all subsequent requires are all the same obj (great for perf, bad for security). Knowing this, you can just modify the cached version in memory. You need to remember to set it back or latter tests will fail due to the corrupted version. Check out my code for an example.

test/index.js
```diff
@@ -0,0 +1,67 @@
...
+
+            Version.register.attributes = {
+                name: 'Fakey Mc Fakerson'
+            };
+
+            lib.init(function (err, server) {
+
+                expect(err).to.equal('I like to break stuff');
+                Version.register = register;
+                server.stop(done);
```
mikejerome: When I tried server.stop(done) here I got an error can't call stop of undefined so I just called done()

TheAlphaNerd: quite strange.

lib/start.js
```diff
@@ -0,0 +1,9 @@
+var Hoek = require('hoek');
+
+var lib = require('./lib');
```
hueniverse: `Lab`

package.json
hueniverse: Doesn't really matter in this case. True for real modules.

test/index.js
```diff
@@ -0,0 +1,57 @@
...
+var Code = require('code');
+var Lab = require('lab');
+var lib = require('../lib');
+var Version = require('../lib/version');
+
+
+// Declare internals
+
+var internals = {
+    port: 1337
```
hueniverse: Never use fixed ports in tests. Always `0` or the default.

test/index.js
```diff
@@ -0,0 +1,57 @@
...
+// Test shortcuts
+
+var lab = exports.lab = Lab.script();
+var describe = lab.describe;
+var it = lab.it;
+var expect = Code.expect;
+
+describe('lib', function() {
+
+    describe('.init()', function () {
```
hueniverse: just `'init()'`.

TheAlphaNerd:
Fixed all requested style issues.

test/index.js
```diff
@@ -0,0 +1,56 @@
...
+describe('Lib', function() {
+
+    describe('init()', { parallel: false }, function () {
+
+        it('should properly handle an error if the plugin does not load correctly', function (done) {
+
+            var register = Version.register;
+
+            Version.register = function (server, options, next) {
+                next('I like to break stuff');
```
hueniverse: Do not return string as error.

PR #82, opened 2015-03-25
stuck at the callback part

roelof1967:
Im stuck in making the init function work on a port and a callback function.  I hope someone can tell me what I did wrong.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
  node_js:
-   - "0.10"
-   - "0.12"
-   - "io.js"  
+  - 0.10
+  - 0.12
+  - io.js  
  
    
    
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: node_js
+ node_js:
+   - "0.10"
+   - "0.12"
+   - "io.js"  
+ 
+   
+   
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 language: node_js
- node_js:
+
+node_js:
   - 0.10
   - 0.12
-  - io.js  
+  - iojs
+
  
    
    
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,19 +10,22 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+internals.init = function (port, function (err) {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    module.exports = server;
+    server.connection({ port: 3000 });
     server.register(Version, function (err) {
 
         Hoek.assert(!err, err);
+
         server.start(function (err) {
 
             Hoek.assert(!err, err);
             console.log('Server started at: ' + server.info.uri);
         });
     });
-};
+});
 
-internals.init();
+internals.init() ;
+
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,30 +10,25 @@ var Version = require('./version');
 var internals = {};
 
 
-exports.init = function (err, data) {
+exports.init = function (data, callback) {
     
-    if (err) { 
-      callback(err, null )
-    }
-
     var server = new Hapi.Server();
     server.connection({ port: Number(data) || 8000 });
     server.register(Version, function (err) {
 
         if (err) {
-            callback(err, null); 
+            return callback(err, server); 
         }
 
         server.start(function (err) {
 
-            if (err) {
-                callback(err, null); 
+           return callback(err, server);
             
-            }
-            console.log('Server started at: ' + server.info.uri);
-        });
-    });
-};
+            });
+                     
+      });
+   };
 
-internals.init() ;
+
+
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,22 +10,30 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function (port, function (err) {
+exports.init = function (err, data) {
+    
+    if (err) { 
+      callback(err, null )
+    }
 
     var server = new Hapi.Server();
-    module.exports = server;
-    server.connection({ port: 3000 });
+    server.connection({ port: Number(data) || 8000 });
     server.register(Version, function (err) {
 
-        Hoek.assert(!err, err);
+        if (err) {
+            callback(err, null); 
+        }
 
         server.start(function (err) {
 
-            Hoek.assert(!err, err);
+            if (err) {
+                callback(err, null); 
+            
+            }
             console.log('Server started at: ' + server.info.uri);
         });
     });
-});
+};
 
 internals.init() ;
 
--- a/lib/version.js
+++ b/lib/version.js
@@ -2,7 +2,6 @@
 
 var Pkg = require('../package.json');
 
-
 // Declare internals
 
 var internals = {
--- a/package.json
+++ b/package.json
@@ -19,10 +19,15 @@
   "homepage": "https://github.com/hueniverse/hueniversity",
   "dependencies": {
     "hapi": "8.x.x",
-    "hoek": "2.x.x"
+    "hoek": "2.x.x",
+    "code": "^1.3.0"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "lab -a code -t 100 -L",
+    "test-cov-html": "lab -a code -r html -o coverage.html",
     "start": "node lib/index.js"
+  },
+  "devDependencies": {
+    "lab": "^5.5.1"
   }
 }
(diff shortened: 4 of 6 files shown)


simon-p-r:
I think you are calling init without a callback so I think this will fail?

roelof1967:
I think I did here :  internals.init = function (port, function (err) {

simon-p-r:
See line 30 of lib/index it calls function without callback

roelof1967:
you are right. Or schould I make there the port and callback part instead of where I did it  or must I do it on both parts ?

TheAlphaNerd:
The second argument in the unit function should simply be callback. You
then cam return and execute that function whenever there is an error,  or
whenever you are ready to return you payload.

Remember that the style for node callbacks is to always make the first
argument the error, followed by return values. I'd there is no error then
you return make the first value null

simon-p-r:
hey @TheAlphaNerd  can you please write a snippet to show us?

roelof1967:
@TheAlphaNerd  I know.
Normally I would use something like this :  init( port, function end(err, data) 
Is that also what you mean ?

TheAlphaNerd:
Nope... I'm on a phone right now, will write in line code review in a
minute when I get to my computer

simon-p-r:
Ok I know when the phone rings you gotta answer!

roelof1967:
@simon-p-r  Would yoy be so kind not to haijck my topic so we can on topic

TheAlphaNerd:
Hehe... I actually was writing from my phone

@roelof1967 we are all in this together here.  Code reviewing coming inline now

simon-p-r:
sorry @roelof1967 I am interested to know the answer also!

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: node_js
+ node_js:
+   - "0.10"
+   - "0.12"
+   - "io.js"  
+ 
+   
+
```
TheAlphaNerd: Too many white spaces! (Should only be one extra)

lib/index.js
```diff
@@ -10,19 +10,22 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+internals.init = function (port, function (err) {
```
TheAlphaNerd: call call back is passed into a function as such

```js
var example = function (data, callback) {
  var payload = data + 1;
  return callback(null, payload);
}

example(9000, function (err, payload) {
  console.log('OMG their power level is: ' + payload);
}
```

TheAlphaNerd: init should also no longer be in the internals object.  Init is the function that we want to export. You will get access to the server through the callback.

```
exports.init = ...
```

lib/index.js
```diff
@@ -10,19 +10,22 @@ var Version = require('./version');
...
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    module.exports = server;
```
TheAlphaNerd: This is not how you expose the server to the lower module  The point of using a callback to return the data is to ensure that the server has properly loaded before you do anything with it.

lib/index.js
```diff
@@ -10,19 +10,22 @@ var Version = require('./version');
...
+    server.connection({ port: 3000 });
```
TheAlphaNerd: The instruction stated that we should be using the port that was provided to the function OR 8000.  Don't know where 3000 came from

lib/index.js
```diff
@@ -10,19 +10,22 @@ var Version = require('./version');
...
     server.register(Version, function (err) {
 
         Hoek.assert(!err, err);
```
TheAlphaNerd: Hoek should no longer be used here, rather we should be checking for an error and returning the execution of the callback if there is one.

lib/index.js
```diff
@@ -10,19 +10,22 @@ var Version = require('./version');
...
+
         server.start(function (err) {
 
             Hoek.assert(!err, err);
```
TheAlphaNerd: See above regarding Hoek assertion

lib/index.js
```diff
@@ -10,19 +10,22 @@ var Version = require('./version');
...
             console.log('Server started at: ' + server.info.uri);
```
TheAlphaNerd: This is where you should be executing the callback to return the server.  The above assertion and error check is not actually necessary as we can send both the value of error and server up the stack here.

lib/index.js
```diff
@@ -10,19 +10,22 @@ var Version = require('./version');
...
         });
     });
-};
+});
+
+internals.init() ;
```
TheAlphaNerd: internals.init() should not be called in lib.  You should make a seperate file at the base of the project called ```start.js``` that requires in this module to execute it.

lib/index.js
roelof1967: I need to change it because Im working on nitrious.io where Im only allowed to use 3000. On final submission I will change it back

TheAlphaNerd: You should be able to set it up properly and pass in appropriate port.

roelof1967:
@TheAlphaNerd  I changed on what I thougt you would mean but when I do nmp test I see this error message :  TypeError: Object #<Object> has no method 'init'

lib/index.js
```diff
@@ -10,19 +10,30 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = function (err, data) {
```
TheAlphaNerd: second argument should be ```callback``` not ```data```

lib/index.js
```diff
@@ -10,19 +10,30 @@ var Version = require('./version');
...
+    
+    if (err) { 
+      callback(err, null )
```
TheAlphaNerd: no need to give a second argument, it will be undefined by default.
Also make sure you return callbacks

lib/index.js
TheAlphaNerd: you are confusing the arguments for you function with the arguments expected by the 

In javascript functions are first class. This means functions can be pass to other functions as arguments (a very powerful feature).

lib/index.js
TheAlphaNerd: also no need to error check here at all (there should be no err in this scope)

lib/index.js
```diff
@@ -10,19 +10,30 @@ var Version = require('./version');
...
+    }
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: Number(data) || 8000 });
     server.register(Version, function (err) {
 
-        Hoek.assert(!err, err);
+        if (err) {
+            callback(err, null);
```
TheAlphaNerd: make sure you return callback, as mentioned above no need to provide the second arg

lib/index.js
```diff
@@ -10,19 +10,30 @@ var Version = require('./version');
...
-        Hoek.assert(!err, err);
+        if (err) {
+            callback(err, null); 
+        }
+
         server.start(function (err) {
 
-            Hoek.assert(!err, err);
+            if (err) {
+                callback(err, null);
```
TheAlphaNerd: no need to check err here.  This is where you return ```callback(err, server)```

lib/index.js
```diff
@@ -10,19 +10,30 @@ var Version = require('./version');
...
+                callback(err, null); 
+            
+            }
             console.log('Server started at: ' + server.info.uri);
         });
     });
 };
 
-internals.init();
+internals.init() ;
```
TheAlphaNerd: you should not be calling init here. This is what is causing you app to throw, as internals no longer has a function called init.

roelof1967:
Thanks, 

I do not see any error messages anymore except this one  :  Object #<Object> has no method 'inject'  

Which I use here : 

~~~

var Lab = require('lab');
var lab = exports.lab = Lab.script();
var server = require('../lib/index.js');
var Code = require('code');

var expect = Code.expect;


lab.test('Testing output on a valid url', function(done) {
    var options = {
        method: 'GET',
        url: '/version'
    };

    server.inject(options, function(response) {
        var result = response.result;

        expect(response.statusCode).to.equal(200);
        expect(result.version).to.equal('0.0.2');

        done();
    });
});

~~~

Roelof

TheAlphaNerd:
lib does not export the server� it exports init.

var lib = require(�../lib�); // no need for index
.
.
.
lib.init(function(err, server) ...

roelof1967:
oke, so I must add exports server = ?? to the end. At some way I need to export the server so my test are working fine . That is why I had exports server = server in my code

TheAlphaNerd:
nope� the callback from init will return the server

TheAlphaNerd:
if you are really lost you might find it useful to look at some of the other pull requests.  Then step away from the computer for a bit so you don�t directly copy the code ;P

PR #83, opened 2015-03-25
Assignment #3: Test Coverage

towns:
I think this is mostly right. Looking forward to any feedback. Thanks for doing this class. Its been fun so far!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,10 @@
+test:
+	@node node_modules/lab/bin/lab -a code
+test-cov:
+	@node node_modules/lab/bin/lab -a code -t 100 -L
+test-cov-html:
+	@node node_modules/lab/bin/lab -a code -r html -o coverage.html
+changelog:
+	@mdchangelog --no-prologue --no-orphan-issues --overwrite --order-milestones semver --order-issues closed_at --dependents --timeout 120000
+
+.PHONY: test test-cov test-cov-html changelog
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,5 @@ test-cov:
 	@node node_modules/lab/bin/lab -a code -t 100 -L
 test-cov-html:
 	@node node_modules/lab/bin/lab -a code -r html -o coverage.html
-changelog:
-	@mdchangelog --no-prologue --no-orphan-issues --overwrite --order-milestones semver --order-issues closed_at --dependents --timeout 120000
 
-.PHONY: test test-cov test-cov-html changelog
+.PHONY: test test-cov test-cov-html
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,19 +10,20 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = function (portToUse, callback) {
+    if (typeof portToUse === 'function') {
+        callback = portToUse;
+        portToUse = undefined;
+    }
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: portToUse || 8000 });
     server.register(Version, function (err) {
-
         Hoek.assert(!err, err);
+
         server.start(function (err) {
 
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+            return callback(err, server);
         });
     });
 };
-
-internals.init();
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,8 +9,8 @@ var Version = require('./version');
 
 var internals = {};
 
+exports.init = function(portToUse, callback) {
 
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -0,0 +1,10 @@
+test:
+	@node node_modules/lab/bin/lab -a code
+test-cov:
+	@node node_modules/lab/bin/lab -a code -t 100 -L
+test-cov-html:
+	@node node_modules/lab/bin/lab -a code -r html -o coverage.html
+changelog:
```
AdriVanHoudt: not part of the assignment :p

towns: hmm, I must be confused. I thought I was satisfying this requirement:

Add the standard hapi.js Makefile to make it easy to generate other types of reports (e.g. HTML coverage report).

AdriVanHoudt: he meant like your test-cov-html, you can also generate the report in other forms, he didn't say anything about changelog generation

lib/index.js
```diff
@@ -10,19 +10,20 @@ var Version = require('./version');
...
+        callback = portToUse;
+        portToUse = undefined;
+    }
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: portToUse || 8000 });
     server.register(Version, function (err) {
-
         Hoek.assert(!err, err);
```
TheAlphaNerd: no need to assert anymore, this will cause your app to throw if there is an error

TheAlphaNerd: this error should be handled via the callback

lib/index.js
```diff
@@ -10,19 +10,20 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = function (portToUse, callback) {
+    if (typeof portToUse === 'function') {
```
TheAlphaNerd: space at beginning of function

lib/start.js
```diff
@@ -0,0 +1,8 @@
+Hoek = require('hoek');
+
+Server = require('./index');
+
+Server.init(8000, function(err, server) {
+    Hoek.assert(!err, err);
```
TheAlphaNerd: missing space at begining of function

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "hueniversity",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "description": "Community learning experiment",
-  "main": "lib/index.js",
+  "main": "lib/start.js",
```
TheAlphaNerd: start should be living in the base of the project not in lib.  It is a consumer of lib

test/index.js
```diff
@@ -0,0 +1,31 @@
...
+var lab = exports.lab = Lab.script();
+
+var Server = require('../lib');
+
+
+lab.test('init with specific port', function (done) {
+
+    Server.init(5000, function(err, server) {
+
+        Hoek.assert(!err, err);
```
TheAlphaNerd: you should be using Code to assert not Hoek

lib/index.js
AdriVanHoudt: unless no callback is given right?

lib/start.js
```diff
@@ -0,0 +1,8 @@
+Hoek = require('hoek');
```
simon-p-r: No variable declared?

lib/start.js
```diff
@@ -0,0 +1,8 @@
...
+
+Server = require('./index');
```
simon-p-r: Same as above, no var declared

towns:
The comment above says that I have merge conflicts that must be resolved. Any tips on what I need to do here? It looks like someone who has write access to the repo has to do it. I obviously don't.

AdriVanHoudt:
@towns you can solve this manually merge master into your branch and resolve the conflicts locally

lib/index.js
```diff
@@ -9,20 +9,20 @@ var Version = require('./version');
 
 var internals = {};
 
+exports.init = function(portToUse, callback) {
```
hueniverse: `port` is more common name.

lib/index.js
```diff
@@ -9,20 +9,20 @@ var Version = require('./version');
...
 
-internals.init = function () {
+    if (typeof portToUse === 'function') {
```
hueniverse: This is not in the assignment.

lib/version.js
```diff
@@ -1,23 +1,23 @@
...
 var internals = {
     response: {
-        version: require('../package.json').version
+        version: Package.version
     }
 };
 
 
-exports.register = function (server, options, next) {
+exports.register = function(server, options, next) {
```
hueniverse: Purposely changing the style?

PR #86, opened 2015-03-26
[Assignment3] lab, code, travis CI and more ....

zoe-1:
Acknowledgements:
Myles Borins TheAlphaNerd assignment-3
was very helpful in completing this assignment.
His code relating to testing: inject, and assertians was helpful.
https://github.com/TheAlphaNerd/hueniversity/tree/assignment-3

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,4 +31,4 @@ help:
 	@echo "  cov-mac	-- run coverage tests output to html file and open in Safari"
 	@echo "  						    "
 
-.PHONY: all test clean
+.PHONY: start install test cov cov-html cov-mac help 
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ cov-html:
 	npm run test-coverage-html
 
 cov-mac:
-	npm run test-coverage
+	npm run test-coverage-html
 	open -a Safari ./test/coverage.html
 
 help:
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,19 @@
+
+start:
+	npm start 
+
+install:
+	npm install
+
+lab:
+	npm test 
+
+lab-cov:
+	npm run test-cov-html
+
+lab-cov-mac:
+	npm run test-cov-html
+	open -a Safari ./test/coverage.html
+
+
+	
--- a/lib/index.js
+++ b/lib/index.js
@@ -4,12 +4,28 @@ var Version = require('./version');
 
 exports.init = function (port, next) {
 
+    var boom = null;
+    if (typeof(port) == 'function') {
+        // Was port number forgotten?
+        next = port;
+        port = 0; 
+        boom = new Error('Error no port submitted');
+    } 
+    else if (typeof(port) != 'number') {
+        // Ensure port is a number?
+        port = 0; 
+        boom = new Error('Error port is not valid number');
+    }
+
     var server = new Hapi.Server();
     server.connection({ port: port });
     server.register(Version, function (err) {
 
         if (err){
             return next(err, server);
+        }
+        else if (boom){
+            return next(boom, server);
         }
 
         server.start(function (err) {
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,6 +9,7 @@ var Version = require('./version');
 
 var internals = {};
 
+exports.Version = Version;
 
 // Initialization
 
@@ -26,7 +27,9 @@ exports.init = function (port, cb) {
 
     server.register(Version, function (err) {
 
-        Hoek.assert(!err, err);
+        if(err){
+            return cb(err, server); 
+        }
 
         server.start(function (err) {
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,19 +10,30 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+// Initialization
+
+exports.init = function (port, cb) {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+
+    if (!port) {
+        port = 8000;
+    }
+
+
+    server.connection({ port: port });
+
+
     server.register(Version, function (err) {
 
         Hoek.assert(!err, err);
+
         server.start(function (err) {
 
             Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+            return cb(err, server);
         });
     });
 };
 
-internals.init();
+
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,7 +9,6 @@ var Version = require('./version');
 
 var internals = {};
 
-exports.Version = Version;
 
 // Initialization
 
@@ -17,9 +16,7 @@ exports.init = function (port, cb) {
 
     var server = new Hapi.Server();
 
-    if (!port) {
-        port = 8000;
-    }
+    port = port || 8000;
 
 
     server.connection({ port: port });
--- a/lib/index.js
+++ b/lib/index.js
@@ -4,16 +4,18 @@ var Hapi = require('hapi');
 var Hoek = require('hoek');
 var Version = require('./version');
 
+
 // Declare internals
 
 var internals = {};
+
 
 internals.init = function () {
 
     var server = new Hapi.Server();
     server.connection({ port: 8000 });
     server.register(Version, function (err) {
-      
+
         Hoek.assert(!err, err);
         server.start(function (err) {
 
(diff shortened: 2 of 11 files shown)


Makefile
```diff
@@ -0,0 +1,19 @@
...
+
+lab:
+	npm test 
+
+lab-cov:
+	npm run test-cov-html
+
+lab-cov-mac:
+	npm run test-cov-html
+	open -a Safari ./test/coverage.html
```
AdriVanHoudt: what if I don't have safari? :D

zoe-1: Just use: 
  make lab-cov

Then, run your own command to view the coverage.html file. 

make lab-cov-mac command is just a convenience for the mac user. 
I got sick of typing open -a Safari ./test/coverage.html over and over again. :-)

AdriVanHoudt: didn't see the `-mac` :P but this not required for this repo (although i get where you are coming from) also +1 for pointing to npm :D

lib/index.js
```diff
@@ -10,19 +10,30 @@ var Version = require('./version');
...
+        port = 8000;
+    }
+
+
+    server.connection({ port: port });
+
+
     server.register(Version, function (err) {
 
         Hoek.assert(!err, err);
```
TheAlphaNerd: I don't believe we are supposed to use Hoek to assert here.  This would cause the stack to asplode if there is an error and a callback was provided (rather than executing cb with error for handling)

zoe-1: OK. Thanks for the tip. I will make adjustments.

TheAlphaNerd: If you are writing a plugin or module that is going to be consumed by someone else you don't know what they want to do with that error.  Very likely they don't want that error to make their entire application asplode.

I tend to try and write "throw safe" code, passing errors up the callback chain.  Now in theory you should be able to do 

"Try / catch" but that isn't terribly async.

start.js
```diff
@@ -0,0 +1,14 @@
...
+var Hoek = require('hoek');
+var index = require('./lib/index');
+
+
+index.init(null, function (err, server) {
+
+    Hoek.assert(!err, err);
+
+    console.log('Server started at: '+ server.info.uri);
+    console.log('Server port is: '+ server.info.port);
```
chyld: you could combine both console.logs together

start.js
```diff
@@ -0,0 +1,14 @@
+
+var Hoek = require('hoek');
+var index = require('./lib/index');
```
chyld: index is a module, use capital, Index

lib/index.js
```diff
@@ -9,20 +9,33 @@ var Version = require('./version');
 
 var internals = {};
 
+exports.Version = Version;
```
chyld: the only reason you're exporting Version is for a test. Never export something you wouldn't normally export, just to increase code coverage.

lib/index.js
```diff
@@ -9,20 +9,33 @@ var Version = require('./version');
...
 
-internals.init = function () {
+// Initialization
+
+exports.init = function (port, cb) {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+
+    if (!port) {
```
chyld: port = port || 8000

test/version.js
```diff
@@ -0,0 +1,38 @@
...
+var expect = Code.expect;
+
+
+describe('Version', function() {
+
+    it('Return correct version data', function(done){
+
+        Server.init(null,function(err, server){
+
+            Hoek.assert(!err, err);
```
chyld: why are you using Hoek, just make your assertions - they should either pass or fail

zoe-1:
@chyld Thank you for your comments above.  They helped alot. 
I made your recommended changes. Plus, I made a solution
to get 100% coverage without exporting data just to pass tests.  
Please check it out and let me know what you think.  Thanks again for the help!

lib/index.js
```diff
@@ -10,19 +10,29 @@ var Version = require('./version');
 var internals = {};
```
chyld: you're not using internals

chyld:
Good call on the testing plugin failure. I'm not at 100% without having to export something I really didn't want to export.

zoe-1:
@chyld     :-)

Makefile
```diff
@@ -0,0 +1,34 @@
...
+	@echo "  						    "
+	@echo "  command list below:"
+	@echo "  start		-- start node server."
+	@echo "  test 	 	-- run tests"
+	@echo "  cov 		-- run coverage tests output to console."
+	@echo "  cov-html	-- run coverage tests output to html file."
+	@echo "  cov-mac	-- run coverage tests output to html file and open in Safari"
+	@echo "  						    "
+
+.PHONY: all test clean
```
zerok: You should probably also add your other targets to the .PHONY list as this marks them as not being file-related. You can find some details on this here: http://stackoverflow.com/a/2145605/22312

test/version.js
```diff
@@ -0,0 +1,39 @@
...
+describe('Version plugin', function(){
+
+    it('-- ensure version data is correct',function(done){
+    
+        Server.init(null,function(err, server){
+
+            expect(err).to.be.undefined();
+
+
+            expect(server.info.port).to.equal(8000);
```
zerok: IMO this assertion is not really necessary as you are already checking the default port in another test-case :)

zoe-1:
@zerok Thank you for your recommendations.  
I just made the changes.

Makefile
```diff
@@ -0,0 +1,34 @@
...
+	@echo "  						    "
+	@echo "  command list below:"
+	@echo "  start		-- start node server."
+	@echo "  test 	 	-- run tests"
+	@echo "  cov 		-- run coverage tests output to console."
+	@echo "  cov-html	-- run coverage tests output to html file."
+	@echo "  cov-mac	-- run coverage tests output to html file and open in Safari"
+	@echo "  						    "
+
+.PHONY: start install test cov cov-html cov-mac help
```
hueniverse: This is one fancy makefile... :-)

zoe-1:
:-) 
I like having a help command to remind me of the commands I put in there. 
It is probably a bit too much, but it is what it is...

AdriVanHoudt:
Is it possible to automate that? otherwise you have to update the help command every time you add a new command

zoe-1:
@AdriVanHoudt automating the creation of the help command is an interesting thought. 
But I do not think the benefits of automating here would be worth the effort required to write the script to do it.

AdriVanHoudt:
@zoe-1 yeah probably not ^^

EvanMarkHopkins:
what about a cli tool for generating both scripts and a makefile, it could prompt for a *help* description after each script you enter, and even resolve the paths in the makefile :astonished:

AdriVanHoudt:
@EvanMarkHopkins I can see value in a tool to create a makefile from the npm scripts

EvanMarkHopkins:
I can see that, then you could just have a *help* script, it should work both ways though

EvanMarkHopkins:
That's good, maybe have a flag which resolved the module paths if you wanted a standalone makefile

zoe-1:
@EvanMarkHopkins @AdriVanHoudt  your thoughts about automating tasks are interesting.
However, when I think automating things I consider automating tasks that I spend a lot of time doing.
So, in the long wrong automating the task makes me more productive.
I do not spend a lot of time writing Makefile or npm commands.
Usually each project I work on has about 5-10 commands in the package.json file or 
Makefile.  In my opinion, writing a script to automate the creation of 5-10 commands per project is not worth the time and effort needed to make the script.

If I was going to automate something, I would automate the creation of a basic hapijs project. 
It would be a command like: "hapi new <project_name>" and this would generate a 
skeleton for a basic hapijs app etc.  That would save me and others time. 

npm package shelljs -- is nice for using node.js to run BASH commands.
https://github.com/arturadib/shelljs
Check it out.

AdriVanHoudt:
@zoe-1 there is a hapi generator https://github.com/jedireza/generator-hapi-style ;)
and yeah the effort may not be worth it
and you can't expect every windows user to install that :D

TheAlphaNerd:
tools like yargs and commander will generate your help files for you, although this is just for generic node based CLI's, not for a make file.  I believe yargs also gives you free tab completion o0o0o

zoe-1:
@AdriVanHoudt @TheAlphaNerd 
Thank you for the recommendations. :-)

EvanMarkHopkins:
@zoe-1 is nice, but does it not interpret bash and makefiles directly? it would be very nice if one could transpile these to shell.js :smile:

PR #87, opened 2015-03-27
97% coverage lab test and travis

lvegerano:
I would appreciate some help as the route test fails due to plugin error test. The overwritten ```Version``` plugin propagated to the last test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,8 @@
+test:
+	@node node_modules/lab/bin/lab -a code
+test-cov:
+	@node node_modules/lab/bin/lab -a code -t 100 -L
+test-cov-html:
+	@node node_modules/lab/bin/lab -a code -r html -o coverage.html
+
+.PHONY: test test-cov test-cov-html changelog
--- a/Makefile
+++ b/Makefile
@@ -4,5 +4,3 @@ test-cov:
 	@node node_modules/lab/bin/lab -a code -t 100 -L
 test-cov-html:
 	@node node_modules/lab/bin/lab -a code -r html -o coverage.html
-
-.PHONY: test test-cov test-cov-html changelog
--- a/lib/index.js
+++ b/lib/index.js
@@ -5,24 +5,24 @@ var Hoek = require('hoek');
 var Version = require('./version');
 
 
-// Declare internals
+exports.init = function (port, callback) {
 
-var internals = {};
-
-
-internals.init = function () {
+    if (typeof port === 'function') {
+        callback = port;
+        port = 8000;
+    }
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port });
     server.register(Version, function (err) {
 
-        Hoek.assert(!err, err);
+        if (err) {
+            return callback(err);
+        }
+
         server.start(function (err) {
 
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+            return callback(err, server);
         });
     });
 };
-
-internals.init();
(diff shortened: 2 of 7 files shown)


EvanMarkHopkins:
I don't have a ton of knowledge on the subject, and i may misuse a few terms here... but I'm guessing every time you require a module, it's all pointing to the same underlying object. So by giving it a new register property, you're *shadowing* the other references (requires to that module) made further down the * call stack*. So the Version in your last plugin test is still using the function that returns a new error. To illustrate this, you can move your `plugin error test` to the very end, and everything works; But you might want to wait for someone else who has a more elegant solution.

EvanMarkHopkins:
IMO, If the test requires something as aggressive as this, you probably need to rethink your design, and maybe provide yourself a way of adding routes on the fly, or a configuration that adds failing routes under an environment, but that might be over kill for the assignment (providing too much of what wasn't specified). So perhaps for the time being, you'd might be better off simplifying things a little.

zoe-1:
@Ivegerano I was just going to clone your project and begin to debug it. 
But, when I looked at your repository noticed that your working on master. 
Checkout hueniversity guide to working with  git at:
https://github.com/hueniverse/hueniversity/blob/master/guides/git.md

That will make it easier for others to help with your code.  
Plus, in the future, it will help you easily access your code from all various assignments 
rather than just the most recent one.

zoe-1:
@lvegerano 
I cloned your project and worked on it a while. 
It now works with 100% coverage.  
I made  a patch to fix your test/index.js file. Link below:
https://github.com/zoe-1/hueniversity/blob/patch-2/test/index.js
Main issue was the "server handles plugin errors" test.

Please make the changes that I commented on and let me know what happens. 
@TheAlphaNerd, @chyld and @idanwe are all good sources to look at if you want to explore more.
Regards.

lvegerano:
@zoe-1 @EvanMarkHopkins Thank you all for your feedback

Makefile
```diff
@@ -0,0 +1,8 @@
+test:
+	@node node_modules/lab/bin/lab -a code
+test-cov:
+	@node node_modules/lab/bin/lab -a code -t 100 -L
+test-cov-html:
+	@node node_modules/lab/bin/lab -a code -r html -o coverage.html
+
+.PHONY: test test-cov test-cov-html changelog
```
zerok: I don't see a changelog target here :)

lib/index.js
```diff
@@ -5,24 +5,24 @@ var Hoek = require('hoek');
...
 
 
-// Declare internals
+exports.init = function (port, callback) {
 
-var internals = {};
-
-
-internals.init = function () {
+    if (typeof port === 'function') {
```
hueniverse: No in the assignment.

test/index.js
```diff
@@ -0,0 +1,54 @@
+var Lab = require('lab');
+var Code = require('code');
+var Server = require('../lib');
+
+
+var lab = exports.lab = Lab.script();
+
+lab.experiment('Server', function () {
```
hueniverse: Better off using the shortcuts boilerplate from hapi tests.

test/version.js
```diff
@@ -0,0 +1,29 @@
...
+var lab = exports.lab = Lab.script();
+var expect = Code.expect;
+
+lab.experiment('Version Plugin', function () {
+
+    lab.test('server returns plugin version', function (done) {
+
+        Server.init(function (err, server) {
+
+            var request = {
```
hueniverse: `inject()` reads better for GET requests when the request is just the path strings.

PR #88, opened 2015-03-27
Final submission : 100 % coverage 

roelof1967:
Hello, 

Lab says there is still no coverage on line 16 of index.js. 
I do not agree with it because I wrote two test which test if a server starts up on port 8000 (standard) and on port 5000. And the test are passed.

Anyone a tip how to solve this ?


test/index.js
```diff
@@ -0,0 +1,63 @@
...
+var lab = exports.lab = Lab.script();
+var describe = lab.experiment;
+var expect = Code.expect;
+var it = lab.test;
+
+describe('Server', function() {
+   
+    it('starts server on port 8000 by default', function (done) {
+
+        Server.init(8000, function (err, server) {
```
EvanMarkHopkins: 8000 satisfies the first condition, try passing null or 0 instead.

roelof1967:
@EvanMarkHopkins  Thanks, one small change and 100 % coverage. Now hope that the experts are happy too.

lib/index.js
```diff
@@ -10,19 +10,25 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = function (data, callback) {
```
chyld: the variable name data is generic, use port

roelof1967:
@chyld  thanks for the feedback, changed it already

.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: node_js
+
+node_js:
+  - 0.10
+  - 0.12
+  - iojs
+
```
idanwe: Remove empty lines

lib/index.js
```diff
@@ -10,19 +10,25 @@ var Version = require('./version');
...
+        }
+
         server.start(function (err) {
 
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
-        });
-    });
-};
+           return callback(err, server);
```
idanwe: Fix the indentations and remove the empty lines after `});`

lib/index.js
```diff
@@ -10,19 +10,25 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = function (port, callback) {
```
roelof1967: oke, 
Can you explain what you mean here ?

idanwe: There is two possibilities for the function arguments, `(callback)` or `(port, callback)`

And there are several solutions for this:
 * Type checking of the arguments - you can look at [my PR](https://github.com/idanwe/hueniversity/blob/assignment3/lib/index.js#L16-L19)
 * Arguments length checking - [hapi code](https://github.com/hapijs/hapi/blob/master/lib/server.js#L205-L206)

roelof1967: Thanks, but as soon as I added this : 

~~~

if (typeof port === 'function') {
        callback = port;
        port = internals.port;
    }

~~~  

Then all the test of index are failing. I have then a coverage of 93%.

lib/index.js
```diff
@@ -4,25 +4,29 @@ var Hapi = require('hapi');
...
 
-var internals = {};
+var internals = {
+    port: 8000
+};
 
+exports.init = function (port, callback) {
 
-internals.init = function () {
+    if (typeof port === 'function') {
```
hueniverse: Not in the assignment.

start.js
```diff
@@ -0,0 +1,9 @@
+var Hoek = require('hoek');
+
+var lib = require('./lib');
```
hueniverse: `Lib`

lib/index.js
roelof1967: oke, a few remarks before you idanwe told me to change it to this. Is this then a good solution :  port = port || 0 ;

start.js
roelof1967: I will change this

idanwe: Sorry for the confusion

lib/index.js
AdriVanHoudt: Hapi will do that for you ;)

start.js
```diff
@@ -0,0 +1,9 @@
+var Hoek = require('hoek');
+
+var Lib = require('./lib');
+
+lib.init(8000, function (err, server) {
```
arb: You sure this works?

.travis.yml
```diff
@@ -0,0 +1,6 @@
+language: node_js
+
+node_js:
+  - 0.10
+  - 0.12
+  - iojs
```
arb: Missing newline.

start.js
roelof1967: Yep. all the test I made were green.

.travis.yml
roelof1967: Where does the newline to be placed ?

AdriVanHoudt: at the end of the file

PR #91, opened 2015-03-28
[Assignment 3] Add tests and externalize server initialization

zerok:


lib/index.js
```diff
@@ -10,19 +9,24 @@ var Version = require('./version');
 var internals = {};
```
chyld: don't need internals or internals.init. you can just delete it. it's not used.

lib/start.js
```diff
@@ -0,0 +1,8 @@
+var Server = require('./index');
```
chyld: start.js should be in the root directory

zerok:
@chyld both corrected. Thanks for the feedback :)

zoe-1:
@zerok checkout @hueniverse comments made in PR #80. 
There are many changes described in that PR that you need to implement. 
Check it out :-)

PR #92, opened 2015-03-28
Assignment3

rutaihwa:
#79  Testing the app, probably tests should be more robust. Not sure about the Makefile, i just got it ast it is from hapi package.


lib/index.js
```diff
@@ -4,25 +4,23 @@ var Hapi = require('hapi');
...
+    }
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port });
     server.register(Version, function (err) {
 
-        Hoek.assert(!err, err);
-        server.start(function (err) {
+	Hoek.assert(!err, err);
```
chyld: don't assert, just pass err back to caller

start.js
```diff
@@ -0,0 +1,10 @@
+// Load modules
+
+var Hoek = require('hoek');
+var lib = require('./lib');
```
chyld: call it Server instead of lib ... it's more descriptive. Lib means nothing.

rutaihwa: @chyld, that makes sense..

lib/index.js
rutaihwa: sweet, let the caller handle errors.

PR #93, opened 2015-03-28
Assignment 3

toboid:
#79

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,7 +1,6 @@
 // Load modules
 
 var Hapi = require('hapi');
-var Hoek = require('hoek');
 var Version = require('./version');
 
 
@@ -10,19 +9,24 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = function (port, callback) {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port });
     server.register(Version, function (err) {
 
-        Hoek.assert(!err, err);
+        if (err) {
+          return callback(err);
+        }
+
         server.start(function (err) {
 
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+            if (err) {
+              return callback(err);
+            }
+
+            return callback(null, server);
         });
     });
 };
 
-internals.init();
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,19 +9,19 @@ var Version = require('./version');
 var internals = {};
 
 
-exports.init = function (port, callback) {
+exports.init = function (port, next) {
 
     var server = new Hapi.Server();
     server.connection({ port: port });
     server.register(Version, function (err) {
 
         if (err) {
-          return callback(err, server);
+          return next(err, server);
         }
 
         server.start(function (err) {
 
-            return callback(err, server);
+            return next(err, server);
         });
     });
 };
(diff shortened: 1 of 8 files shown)


lib/index.js
```diff
@@ -10,19 +9,20 @@ var Version = require('./version');
 var internals = {};
```
chyld: you're not using internals

lib/start.js
```diff
@@ -0,0 +1,16 @@
+// Load modules
+
+var Hoek = require('hoek');
+var Index = require('./');
+
+
+// Declare internals
+
+var internals = {};
```
chyld: not using internals

lib/start.js
```diff
@@ -0,0 +1,16 @@
+// Load modules
```
chyld: start.js should be in the root directory

test/index.js
```diff
@@ -0,0 +1,67 @@
...
+
+describe('Server', function () {
+
+    it('listens on supplied port', function (done) {
+
+        Index.init(8000, function (err, server) {
+
+            expect(err).to.not.exist();
+            expect(server.info.port).to.equal(8000);
+            server.stop();
```
chyld: put done inside stop ... server.stop(done)

test/index.js
```diff
@@ -0,0 +1,67 @@
...
+            var routes = server.connections[0].table();
+            var foundVersion = false;
+            routes.forEach(function (route) {
+                if (route.path === '/version') {
+                    foundVersion = true;
+                }
+            });
+            expect(foundVersion, 'version route not found').to.equal(true);
+            server.stop();
+            done();
```
chyld: done inside stop

toboid: Thanks for feedback, I won't have access a computer for a few days now so won't be able to update this. I appreciate the comments though

lib/index.js
toboid: Yes, I wondered about this and not really comfortable with having a redundant variable there either. However I think this is required by the [style guide](http://hapijs.com/styleguide):
> internals - must be declared as an object at the top of each module immediate following the require section

Also had a quick look at the hapi source code and found examples of the empty variable that is not used in the rest of the file there too e.g. https://github.com/hapijs/hapi/blob/master/lib/index.js.

Be interested to get @hueniverse take on this.

lib/start.js
toboid: Seemed to be different views on this as it wasn't specified in the task, but updated as think makes sense in the root directory

EvanMarkHopkins:
I think it was actually decided that ```start.js``` would go in lib. I will look for the PR

lib/index.js
```diff
@@ -10,19 +9,20 @@ var Version = require('./version');
...
 
 
-internals.init = function () {
+exports.init = function (port, callback) {
```
EvanMarkHopkins: use ```next``` instead of ```callback```, also in #80 (i think)

zoe-1:
@EvanMarkHopkins  is right. start.js goes in lib directory.

zoe-1:
Wow! @EvanMarkHopkins is going beyond the call of duty.  He is giving references to 
PRs for each recommendation.  nice work @EvanMarkHopkins. very kind of you.

EvanMarkHopkins:
@zoe-1 I'm asking them to revert something they just finished *fixing*, the least I can do is validate my claims :smile:

zoe-1:
@EvanMarkHopkins  just wanted to give you a compliment. :-) 
Thank you for doing diligent work. Do not often see people making the effort like that
so I was impressed.

toboid:
thanks for the link, missed that. moving back to lib

lib/index.js
toboid: thanks :smile:

PR #94, opened 2015-03-28
[Assignment 3] 100% Coverage

sfabriece:


lib/start.js
```diff
@@ -0,0 +1,7 @@
+var Index = require('./index');
+var Hoek = require('hoek');
+Index.init(8000, function(err, server){
+    Hoek.assert(!err, err);
+    console.log('Server started at: ' + server.info.uri);
+    return server;
```
chyld: don't return the server

sfabriece:
done :)

EvanMarkHopkins:
I believe the -L flag in lab enforces these styles by default, so you don't need to specify them yourself. Also, the 4th objective specified a makefile, there's nothing wrong with what you have done, but it's worth giving a shot.

sfabriece:
added makefile, can you elaborate on why it is preferable to use a makefile?

lib/start.js
```diff
@@ -0,0 +1,6 @@
+var Index = require('./index');
+var Hoek = require('hoek');
+Index.init(8000, function(err, server){
```
AdriVanHoudt: newline after function and also between modules and the init() function call

package.json
```diff
@@ -1,28 +1,33 @@
...
-  "name": "hueniversity",
-  "version": "0.0.2",
-  "description": "Community learning experiment",
-  "main": "lib/index.js",
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/hueniverse/hueniversity.git"
-  },
-  "keywords": [
+	"name": "hueniversity",
```
AdriVanHoudt: why did you change the indentation?

test/index.js
```diff
@@ -0,0 +1,46 @@
+var Code = require('code');
+var Lab = require('lab');
+var Server = require('../lib');
+var Version = require('../lib/version');
+
+var expect = Code.expect;
+var labi = exports.lab = Lab.script();
+
+labi.experiment('index', function(){
+    labi.test('init starts the server', function(done){
```
AdriVanHoudt: newline after function

test/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+        Server.init(8081, function(err, server){
```
AdriVanHoudt: better to not pass along a defined port since that might interfere with other tests

test/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+    labi.test('should report error if plugin registration fails', function(done){
+
+        var correctVersionPlugin = Version.register;
+        var fakePluginError = 'Fake Plugin Error';
+        Version.register = function(server, options, next){
+
+            return next(fakePluginError);
+        };
+        Version.register.attributes = {name: 'Fake version Plugin' };
```
AdriVanHoudt: space after {

test/version.js
```diff
@@ -0,0 +1,25 @@
...
+
+        Server.init(8000, function(err, server){
+
+            var options = {
+                url: '/version'
+            };
+            server.inject(options, function(res){
+
+                expect(res.result).to.be.an.object();
+                expect(res.result.version).to.equal(Pkg.version);
```
AdriVanHoudt: what if the server responds with an object with an additional (not wanted) key? The test would still pass...

AdriVanHoudt:
@sfabriece because learning and Eran likes it ^^ also @EvanMarkHopkins is right when using -L lab will use predefined rules for Hapi that come along with it

package.json
EvanMarkHopkins: maybe an editor problem? it seems to handle JS pretty well, but perhaps some special attention to JSON is in order. Requiring a final newline in files is another thing to look into.

sfabriece: an editor problem and I forgot to run eslint.

zoe-1:
@sfabriece The Makefile is useful for automating tasks in your project. 
npm commands can execute automated tasks but they can get very long.  Long commands are inconvenient to type over and over.  

For example, npm allows you to run two commands related to executing scripts by default: "npm start" and "npm test".  If you make other custom npm commands in the scripts section, they will fail without "run" preceding the command.
ex) "npm run command_name" // illustrates a command with run before it.

See below command result for npm's default commands that do not need a "run" before them.
npm --help command  // This will list npm's commands (test and start are included here).

Look at: https://github.com/zoe-1/nmax/blob/master/package.json to see a package.json file with a significant amount of package.json scripts commands in them. 
The commands are made in the scripts:{ } section.
In the above linked package.json file there is the below command configured:
"protractor": "protractor test/protractor-conf.js"
To execute this command you must type: "npm run protractor". "npm protractor" without "run" will not work . "npm run protractor" will run tests on the project. The command is a long command to type every time you want to run tests.  The Makefile lets you simplify this.  
Put in your Makefile:
----<br/>
`pro:`
&nbsp;&nbsp;&nbsp;&nbsp;`npm run protractor`<br/>


Then type:<br/>
&nbsp;&nbsp;&nbsp;&nbsp;`make pro`<br/>

And, the long "npm run protractor" command will be executed.  

As a project develops you will need automate tasks like: 
1. changes to the configuration state (live versus development states)
2. execution of tests
3. server start up 
4. and others....(For example, git push commands to different remote repos) 

In this assignment, I believe Eran is showing us how to use the Makefile as a convenient method
to manage and execute our projects most important automated tasks.

sfabriece:
@zoe-1 Thanks for the explanation.

test/index.js
sfabriece: Should this be caught by the linter?

test/index.js
sfabriece: does deep equality test solve this?

AdriVanHoudt: what do you mean? The problem is that when you write other tests and you use the same port there by accident it will cause issues, so better to not use specified ports unless for the test whether init() will use the specified port

test/index.js
AdriVanHoudt: yes but the lint rules are not complete(feel free to add to them https://github.com/hapijs/lab/blob/master/lib/linters/eslint/.eslintrc )

PR #95, opened 2015-03-29
Assignment 3

### LINKED ISSUE #79: 100% coverage

AdriVanHoudt:
closes #79

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: node_js
+node_js:
+- "0.10"
+- "0.12"
+- iojs
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,7 +1,6 @@
 // Load modules
 
 var Hapi = require('hapi');
-var Hoek = require('hoek');
 var Version = require('./version');
 
 
@@ -10,19 +9,19 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = internals.init = function (port, next) {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port });
     server.register(Version, function (err) {
 
-        Hoek.assert(!err, err);
+        if (err) {
+            return next(err);
+        }
+
         server.start(function (err) {
 
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+            return next(err, server);
         });
     });
 };
-
-internals.init();
--- a/test/index.js
+++ b/test/index.js
@@ -1,0 +1,72 @@
+var Code = require('code');
+var Lab = require('lab');
+
+var index = require('../lib/index');
+
+
+// Test shortcuts
+
+var lab = exports.lab = Lab.script();
+var describe = lab.describe;
+var it = lab.it;
+var expect = Code.expect;
+
+
+describe('index.js', function () {
+
+    describe('init()', function () {
+
+        it('Returns a server', function (done) {
+
+            return index.init(null, function (err, server) {
+
+                expect(err, 'error').to.not.exist();
+                expect(server, 'server').to.be.an.object();
+
+                server.stop(done);
+            });
+        });
+
+        it('Returns an error when a plugin fails to register', { parallel: false }, function (done) {
+
+            // Replace the register function to force an error that can be caught by init()
+            // Make sure to backup the old register function and set it back!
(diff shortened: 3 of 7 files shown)


lib/index.js
```diff
@@ -10,19 +9,19 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = internals.init = function (port, callback) {
```
chyld: you're not using internals.init.

lib/index.js
```diff
@@ -10,19 +9,19 @@ var Version = require('./version');
...
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port });
```
chyld: port is optional. you're not checking for undefined here.

EvanMarkHopkins: the assignment never specified optional, and port can handle null and undefined here, it will even take strings like '8000'.

AdriVanHoudt: Idd, port was not specified to be optional and the null/undefined will be handled by Hapi and will resolve to 8000, just like host defaults to localhost

lib/index.js
AdriVanHoudt: True but I thought it was the Hapi style, I have seen it in other Hapi repos
@hueniverse can you clarify if this is wanted or not needed?

lib/index.js
EvanMarkHopkins: it won't resolve to 8000... but the assignment never specified defaulting to 8000 either, it only sugessted setting 8000 in start.js

AdriVanHoudt: my bad, looks like it defaults to 0

EvanMarkHopkins: which is fine, hapi will find an available port for you.

AdriVanHoudt: yeah and for testing with server.inject you don't need a special port

lib/index.js
chyld: it is Hapi style, but you just created a variable that you're not using.

AdriVanHoudt: true but I've noticed that the hapi style prefers to have internals and things in internals when needed even though it might not be needed right now, I really want to hear from @hueniverse for this

EvanMarkHopkins: The source code does do this, but from what I looked at, you're not touching prototype, or calling with **new**. So it's probably not necessary for init to occupy both globals

EvanMarkHopkins: I'm also curious how the **this** would resolve here, does it refer to exports, or internals? Not that you're messing with that. I'm just curious.

EvanMarkHopkins: Unrelated: The [style guide](https://github.com/hapijs/contrib/blob/master/Style.md#callback) does say if you're not using process.nextTick, callback should be ```next```

AdriVanHoudt: I can test the **this** for you tomorrow if you want :P and I'm pretty sure the callback will resolve on nextTick although I'm not super familiar with the workings

EvanMarkHopkins: no, that's okay, it refers to exports, and I don't know JS about nextTick, i assumed it meant called explicitly :grin:

AdriVanHoudt: you were right though, saw Eran comment in @chyld's PR and I found this to explain it fairly easily http://www.quora.com/What-does-process-nextTick-callback-actually-do-in-Node-js

test/index.js
```diff
@@ -0,0 +1,65 @@
+var Code = require('code');
+var Lab = require('lab');
+
+var index = require('../lib/index');
+
+var lab = exports.lab = Lab.script();
+var expect = Code.expect;
+
+lab.experiment('Index.js', function () {
```
hueniverse: `index.js`

.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: node_js
+node_js:
+- "0.10"
+- "0.12"
+- iojs
```
arb: Missing newline.

AdriVanHoudt: is it needed in a yml file?

arb: https://github.com/hapijs/contrib/blob/master/Style.md#newlines

AdriVanHoudt: I assumed that was only for js files, give me a sec to update

AdriVanHoudt: fixed but github doesn't make this outdated :(

PR #96, opened 2015-03-29
Assignment 3

gmertk:


lib/index.js
```diff
@@ -10,19 +9,27 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = internals.init = function (port, callback) {
+
+    if (typeof port === 'function') {
+        callback = port;
+        port = 8000;
+    } else if (port === null || port === undefined) {
```
EvanMarkHopkins: You'll wanna test both sides of this. However, since someone will have to explicitly enter ```undefined```, or omit the callback as well (something you aren't checking for) you might as well drop it. That's entirely up to you though.

EvanMarkHopkins: You could get as picky as you want, check If port [isNaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN), Or if it's an array... But just realize you'll have to write more tests to check **every** scenario.

hueniverse:
Did you notice you are missing coverage?

PR #97, opened 2015-03-30
Assignment 3 commit

simon-p-r:


lib/index.js
```diff
@@ -10,19 +10,24 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = internals.init = function (port, callback) {
```
chyld: you're not using internals. you're creating the property init, but you're not using it - just remove internals. it's not needed here.

simon-p-r:
Ok @chyld, I have removed the internal.init on your recommendation.  Thanks for the feedback!

lib/index.js
```diff
@@ -10,19 +10,24 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = function (port, callback) {
```
EvanMarkHopkins: Replace all instances of ```callback``` with ```next```, the reason behind such naming convensions are described by eran in #80 and in the [style guide](https://github.com/hapijs/contrib/blob/master/Style.md#callback)

simon-p-r:
Ok thanks @EvanMarkHopkins

PR #98, opened 2015-03-30
[Assignment 3] 100% coverage

adrian-tiberius:


lib/index.js
```diff
@@ -10,19 +10,20 @@ var Version = require('./version');
...
 
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+            return callback(err, server);
         });
     });
 };
 
-internals.init();
+exports = module.exports = internals.init;
```
chyld: a much simpler version would be exports.init = function(port, callback){}. and that's it. you're done.

start.js
```diff
@@ -0,0 +1,8 @@
+var Server = require('./lib/index');
```
chyld: or ```var Server = require('./lib')``` index.js is the default file.

AdriVanHoudt: being more specific can't hurt though

lib/index.js
```diff
@@ -10,19 +10,20 @@ var Version = require('./version');
 var internals = {};
```
hussainanjar: You are still requiring Hoek though its not used any more .. made the same mistake :smile:

test/index.js
```diff
@@ -0,0 +1,50 @@
...
+        Server(internals.port, function(err, server) {
+
+            expect(!err).to.equal(true);
+            expect(server.info.port).to.equal(internals.port);
+
+            server.stop(done);
+        });
+    });
+
+    it('test plugin error', function (done) {
```
hussainanjar: Monkey patching code needs { parallel: false } in the `it('test plugin error', { parallel: false }, function (done) {`
See @hueniverse comments made in @chyld PR #80 and the main issue.

hussainanjar:
You need to add new line to almost all your files, why new line ? This may help http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline

PR #102, opened 2015-04-01
Assignment 3: 

gyaresu:
Not complete but submitting anyway. 

Can't figure out why my single test won't run.

Thanks

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -6,11 +6,6 @@ var internals = {}; // Global variable https://gist.github.com/hueniverse/a06f63
 exports.init = function (port, next) {
 
 
-    if (typeof port === 'function') {
-        next = port;
-        port = 8000;
-    }
-
     var server = new Hapi.Server();
     server.connection({ port: port });
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,12 +1,10 @@
 var Hapi = require('hapi');
-// Removing Hoek because we are sending the error back to start.js with the error in the callbck
-//var Hoek = require('hoek');
 var Version = require('./version');
 
 exports.init = function (port, callback) {
 
     if (typeof port === 'function') {
-        cb = port;
+        callback = port;
     }
     else if (port === null) {
         port = 8000;
--- a/lib/index.js
+++ b/lib/index.js
@@ -11,7 +11,6 @@ exports.init = function (port, callback) {
     }
 
     var server = new Hapi.Server();
-
     server.connection({ port: port });
 
     server.register(Version, function (err) {
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,10 +3,11 @@ var Version = require('./version');
 
 var internals = {}; // Global variable https://gist.github.com/hueniverse/a06f6315ea736ed1b46d
 
-exports.init = function (port, callback) {
+exports.init = function (port, next) {
+
 
     if (typeof port === 'function') {
-        callback = port;
+        next = port;
         port = 8000;
     }
 
@@ -16,12 +17,12 @@ exports.init = function (port, callback) {
     server.register(Version, function (err) {
 
         if (err) {
-            return callback(err, server);
+            return next(err, server);
         }
 
         server.start(function (err) {
 
-            return callback(err, server);
+            return next(err, server);
         });
     });
 };
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,26 +1,32 @@
-// Load modules
-
 var Hapi = require('hapi');
-var Hoek = require('hoek');
+// Removing Hoek because we are sending the error back to start.js with the error in the callbck
+//var Hoek = require('hoek');
 var Version = require('./version');
-
-
-// Declare internals
-
-var internals = {};
-
 
 exports.init = function (port, callback) {
 
+    if (typeof port === 'function') {
+        cb = port;
+    }
+    else if (port === null) {
+        port = 8000;
+    }
+
     var server = new Hapi.Server();
-    server.connection({ port: port});
+
+    server.connection({ port: port });
+
     server.register(Version, function (err) {
 
-        Hoek.assert(!err, err);
+        if (err) {
+            return callback(err, server);
+        }
+
         server.start(function (err) {
 
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+            if (err) {
+                return callback(err, server);
+            }
         });
     });
 };
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,10 +10,10 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = function (port, callback) {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port});
     server.register(Version, function (err) {
 
         Hoek.assert(!err, err);
@@ -24,5 +24,3 @@ internals.init = function () {
         });
     });
 };
-
-internals.init();
--- a/lib/index.js
+++ b/lib/index.js
@@ -5,8 +5,6 @@ exports.init = function (port, callback) {
 
     if (typeof port === 'function') {
         callback = port;
-    }
-    else if (port === null) {
         port = 8000;
     }
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -21,9 +21,7 @@ exports.init = function (port, callback) {
 
         server.start(function (err) {
 
-            if (err) {
-                return callback(err, server);
-            }
+            return callback(err, server);
         });
     });
 };
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,6 +1,8 @@
 var Hapi = require('hapi');
 var Version = require('./version');
 
+var internals = {}; // Global variable https://gist.github.com/hueniverse/a06f6315ea736ed1b46d
+
 exports.init = function (port, callback) {
 
     if (typeof port === 'function') {
--- a/lib/start.js
+++ b/lib/start.js
@@ -4,11 +4,9 @@
 var Hoek = require('hoek');
 
 // Requiring './lib' like this assigns the `register` & `init` methods to the variable `lib`.
-
 var lib = require('./index');
 
 // port and the callback are called
-
 lib.init(8000, function (err, server) {
 
 
--- a/lib/start.js
+++ b/lib/start.js
@@ -1,5 +1,5 @@
-// Start.js - Why is this required over just index.js?
-//
+// Q: Start.js - Why is this required over just index.js?
+// A: Because you don't want to expose the automatic server creation as an API.
 var Hoek = require('hoek');
 
 // Requiring './lib' like this assigns the `register` & `init` methods to the variable `lib`.
--- a/lib/start.js
+++ b/lib/start.js
@@ -4,7 +4,7 @@ var Hoek = require('hoek');
 
 // Requiring './lib' like this assigns the `register` & `init` methods to the variable `lib`.
 
-var lib = require('./lib/index');
+var lib = require('./index');
 
 // port and the callback are called
 
--- a/lib/start.js
+++ b/lib/start.js
@@ -10,6 +10,7 @@ var lib = require('./index');
 
 lib.init(8000, function (err, server) {
 
+
     Hoek.assert(!err, err);
     console.log('Server started at: ' + server.info.uri);
 });
--- a/lib/start.js
+++ b/lib/start.js
@@ -1,17 +1,15 @@
-// Start.js - Why is this required over just index.js? 
-// 
+// Start.js - Why is this required over just index.js?
+//
 var Hoek = require('hoek');
 
-// Requiring './lib' like this assigns the `register` & `init` methods to the variable `lib`. 
+// Requiring './lib' like this assigns the `register` & `init` methods to the variable `lib`.
 
 var lib = require('./lib/index');
 
-// port and the callback are called 
+// port and the callback are called
 
 lib.init(8000, function (err, server) {
 
     Hoek.assert(!err, err);
     console.log('Server started at: ' + server.info.uri);
 });
-
-
--- a/lib/start.js
+++ b/lib/start.js
@@ -1,5 +1,6 @@
 // Q: Start.js - Why is this required over just index.js?
 // A: Because you don't want to expose the automatic server creation as an API.
+
 var Hoek = require('hoek');
 
 // Requiring './lib' like this assigns the `register` & `init` methods to the variable `lib`.
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "hoek": "2.x.x"
   },
   "scripts": {
-    "test": "lab",
+    "test": "make test-cov",
     "test-cov": "lab -a code -t 100 -L",
     "test-cov-html": "lab -a code -r html -o coverage.html"
   },
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hueniversity",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "description": "Community learning experiment",
   "main": "lib/index.js",
   "repository": {
@@ -22,7 +22,11 @@
     "hoek": "2.x.x"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "lab -t 100",
     "start": "node lib/index.js"
+  },
+  "devDependencies": {
+    "code": "^1.3.0",
+    "lab": "^5.5.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "hueniversity",
   "version": "0.0.3",
   "description": "Community learning experiment",
-  "main": "start.js",
+  "main": "lib/start.js",
   "repository": {
     "type": "git",
     "url": "https://github.com/hueniverse/hueniversity.git"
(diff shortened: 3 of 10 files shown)


lib/index.js
```diff
@@ -1,28 +1,29 @@
...
-        Hoek.assert(!err, err);
+        if (err) {
+            return callback(err, server);
+        }
+
         server.start(function (err) {
 
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+            if (err) {
```
EvanMarkHopkins: What about when there is no error? :wink:

gyaresu: I put the server.start anonymous function's return into an error checking `if` statement.
And yet that server.start was still running and returning a call to `/version`
So it was only the callback that was failing. 
Which the lab `Server.init` callback was failing on because it wasn't getting a server object back with which to get `server.info.port` from.

Thanks!

EvanMarkHopkins: I dont think server.inject requires a server to be running, so i could see the confusion that would cause you

zoe-1:
Just cloned and ran your project. 
You need to make a test that tests lines 6-10.
%100 requires you to make tests for the port variable being evaluated below.
if (typeof port === 'function') {
        callback = port;
    }
    else if (port === null) {
        port = 8000;
    }

zoe-1:
But, your logic in this if else statement has some problems.

zoe-1:
:-)

gyaresu:
Thanks @zoe-1 & @EvanMarkHopkins you're super helpful.

package.json
```diff
@@ -22,7 +22,12 @@
     "hoek": "2.x.x"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "start": "node lib/index.js"
+    "test": "lab",
```
EvanMarkHopkins: Not a whole lot of people did this, but since you already have the makefile, you have the option of running `make test-cov` and spare yourself the hasle of moving it over.

gyaresu: Cool. So `"test": "make test-cov",` and then I can just `npm test`. Is that what you mean? 

p.s. I didn't even know you could do in-line edits. Sweet.

EvanMarkHopkins: yes, although, the make might have been removed from the assignment, so you dont have to change it, it was just fyi. :smile:

EvanMarkHopkins: and i discover new things about github everyday, for instance i never realized their was a ```split``` diff toggle up at the top of the page, so i just got used to staring at a ton of red lines all day u_ u

lib/index.js
```diff
@@ -1,28 +1,25 @@
...
 
+exports.init = function (port, callback) {
 
-// Declare internals
-
-var internals = {};
-
-
-internals.init = function () {
+    if (typeof port === 'function') {
```
AdriVanHoudt: this requirement was removed from the assignment

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "hueniversity",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "description": "Community learning experiment",
-  "main": "lib/index.js",
+  "main": "start.js",
```
AdriVanHoudt: start.js should be in /lib

package.json
```diff
@@ -22,7 +22,12 @@
     "hoek": "2.x.x"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "start": "node lib/index.js"
+    "test": "make test-cov",
+    "test-cov": "lab -a code -t 100 -L",
+    "test-cov-html": "lab -a code -r html -o coverage.html"
```
AdriVanHoudt: personally I don't like to have 2 places where my test scripts are defined (my solution for example is to define them in package and point my makefile to `npm run test-cov` for example

package.json
```diff
@@ -22,7 +22,12 @@
...
+  },
+  "devDependencies": {
+    "code": "^1.3.0",
```
AdriVanHoudt: better to also use x style notation as in 1.x.x

test/index.js
```diff
@@ -0,0 +1,58 @@
...
+
+describe('Should create a server with default port', function() {
+
+    it('starts the server', function(done) {
+
+        Server.init(8000, function(err, server) {
+
+            expect(err).to.be.undefined();
+            expect(server.info.port).to.equal(8000);
+            server.stop(done); // done is getting passed as a callback?
```
AdriVanHoudt: yes done is passed as a callback, this makes sure the test only stops when the server is stopped

test/index.js
```diff
@@ -0,0 +1,58 @@
...
+        });
+    });
+
+    it('Starts server with default port', function(done) {
+        Server.init(function(err, server) {
```
AdriVanHoudt: newline after function

test/version.js
```diff
@@ -0,0 +1,24 @@
...
+
+var lab = exports.lab = Lab.script();
+var expect = Code.expect;
+var describe = lab.experiment;
+var it = lab.test;
+
+describe('Testing the Version plugin', function() {
+
+    it('Returns the server version', function(done) {
+        Server.init(null, function(err, server) {
```
AdriVanHoudt: newline after function

test/version.js
```diff
@@ -0,0 +1,24 @@
...
+
+            server.inject('/version', function(response) {
+
+                expect(response.statusCode).to.equal(200);
+                expect(response.result.version).to.equal(Pkg.version);
```
AdriVanHoudt: Better to test against `response.result` fully, as in the whole object (use `to.deep.equal` ;)) so when extra keys or something slip in the test will catch it

test/version.js
```diff
@@ -0,0 +1,24 @@
...
+            });
+            server.stop(done);
```
AdriVanHoudt: Only stop the server when the inject is fully done (aka the callback is called)

package.json
gyaresu: By two places do you mean in the "scripts" section and the `Makefile`?
Because this way (by advice of zoe-1) I can just run `npm test`

package.json
gyaresu: Will do, although I'm sure the instructions were ambiguous about the location. I though it implied the root dir. (?)

package.json
gyaresu: Thanks. Confusing though because I installed the dev dependencies via `npm --save-dev blah`.
And while I did find the documentation about X-ranges, I couldn't find any info on setting the npm config file to use X-range by default when installing.

Do you have any advice on this? 

Thanks!

[https://docs.npmjs.com/misc/semver](https://docs.npmjs.com/misc/semver)
#### X-Ranges `1.2.x` `1.X` `1.2.*` `*`
 
Any of `X`, `x`, or `*` may be used to "stand in" for one of the
numeric values in the `[major, minor, patch]` tuple.
 
* `*` := `>=0.0.0` (Any version satisfies)
* `1.x` := `>=1.0.0 <2.0.0` (Matching major version)
* `1.2.x` := `>=1.2.0 <1.3.0` (Matching major and minor versions)
 
A partial version range is treated as an X-Range, so the special
character is in fact optional.
 
* `""` (empty string) := `*` := `>=0.0.0`
* `1` := `1.x.x` := `>=1.0.0 <2.0.0`
* `1.2` := `1.2.x` := `>=1.2.0 <1.3.0`

test/version.js
gyaresu: Just hitting the `code` docs now.
A question on debugging: Do you have any advice on how to look at objects such as `response.result` to see what it's returning from within the tests. 
i.e. I normally just `console.log(response.result);` Do you know a better way?

package.json
AdriVanHoudt: i don't know how to make npm do it by default, i just fix it everytime I install a new package :D  the is `--save-exact` but that only forces it to use `1.2.3` instead of `^1.2.3`

test/version.js
AdriVanHoudt: I mainly use node-theseus in combination with brackets(my ide) but I think one of the better ones (that also can do way more) is https://github.com/node-inspector/node-inspector

gyaresu: `expect(response.result).to.deep.equal({ version: '0.0.3' });`

Hard coding the version seems like the wrong way to do it so I'm looking at how to better parse `Pkg.version` without getting the errors below:

```javascript
⇒  npm test

> hueniversity@0.0.3 test /Users/gyaresu/programming/projects/hueniversity
> make test-cov



  ...x

Failed tests:

  4) Testing the Version plugin Returns the server version:

      actual expected

      {
        "version": "0.0.3"
      }

      Expected { version: '0.0.3' } to equal specified value

      at /Users/gyaresu/programming/projects/hueniversity/test/version.js:20:49


1 of 4 tests failed
Test duration: 66 ms
Assertions count: 7 (verbosity: 1.75)
No global variable leaks detected
Coverage: 100.00%
Linting results: No issues

make: *** [test-cov] Error 1
npm ERR! Test failed.  See above for more details.
```

AdriVanHoudt: I did `expect(response.result).to.deep.equal({ version: Pkg.version });`, should work fine

package.json
AdriVanHoudt: yes, what I did is leave them in package.json so you can run `npm test` and point the Makefile to there as well so the Makefile would just `npm test` for you, you can take a look at my PR if you want

package.json
AdriVanHoudt: I know, but @hueniverse commented it somewhere on a PR

test/version.js
gyaresu: :D just found that myself. Started filling it with strings to compare the output. Noticed my failure wasn't highlighting the "0.0.3" bit.

So then realised that Pkg.version was literally returning "0.0.3", not the whole `{ version: "0.0.3" }`

Thanks @AdriVanHoudt I'll push now. 
`expect(response.result).to.deep.equal({ version: Pkg.version });`

AdriVanHoudt: ah yeah like that :+1:

package.json
gyaresu: And it lists the test descriptions this way. Nice.

```javascript
⇒  npm test

> hueniversity@0.0.3 test /Users/gyaresu/programming/projects/hueniversity
> lab -e test -t 100 -vL -a code

Should create a server with default port
  ✔ 1) starts the server (30 ms)
  ✔ 2) Starts server with default port (6 ms)
  ✔ 3) Version plugin fails to load due to error (3 ms)
Testing the Version plugin
  ✔ 4) Returns the server version (18 ms)


4 tests complete
Test duration: 61 ms
Assertions count: 7 (verbosity: 1.75)
No global variable leaks detected
Coverage: 100.00%
Linting results: No issues
```

gyaresu: @AdriVanHoudt Ok. Confused now. 
`npm test` does indeed complete but `npm test-cov` or `npm test-cov-html` doesn't.

They throw errors for "Cannot find module code".

I cloned your repo and got the same.

Ideas?
[https://github.com/AdriVanHoudt/hueniversity/blob/assignment-3/package.json](https://github.com/AdriVanHoudt/hueniversity/blob/assignment-3/package.json)

AdriVanHoudt: try `npm install code` and run again

gyaresu: It's definitely there.

It's something to do with the way the test runs. 

If you run `npm test` it's fine but even if you run the exact "test" command `ab -e test -t 100 -vL -a code` it comes back with `Error: Cannot find module 'code'`
And running `npm test-cov-html` for example just throws back the `npm` help.

Maybe try it on your own repo?

AdriVanHoudt: I just ran `npm run test-cov-html` so there is something wrong on your end. Try deleting the node_modules folder and run `npm install` again

gyaresu: Ah. Ok, well I'm missing `run` from:
```javascript
test-cov-html:
        npm run test-cov-html
```
BUT it seems that neither of us can run:
`npm test-cov-html` 

It only seems callable from:
`make test-cov-html`

**getting warmer**

gyaresu: Ah so it works if you manually call:
`$ npm run test-cov-html`

IDNKT

EvanMarkHopkins:
I don't see the logic in having a makefile that calls npm scripts @AdriVanHoudt. It makes more sense to use npm as a way of tying all your weird bash scripts and makefiles, etc, into one uniform set of commands that the community is familiar with.

AdriVanHoudt:
@EvanMarkHopkins you would call the makefile from package.json? The reason I do it makefile-->npm is that npm is runable on every os, while make is not installed on windows by default, also it is a standard to have your test scripts in npm. To me the makefile is a nice extra but should not be the default

package.json
AdriVanHoudt: `npm test` is a short for `npm run test` for easiness. Other than `nmp start`(which is also a shortcut) you always need to use `npm run [your command]`

EvanMarkHopkins:
I agree with putting it all, or as much as possible into package.js, and when it starts gettin too much to handle, then maybe a build.sh, or makefile would be preferable

AdriVanHoudt:
@EvanMarkHopkins what is the pro of using a makefile that you can't do with npm?

EvanMarkHopkins:
I don't play with make very much, I think bash scripts certainly add more flexibility. I think since he already had the makefile from hapi it would have been easier to just use it. My main pain point is when I start getting really long script lines with && and calls npm start itself, that I start thinking maybe json wasn't the optimal choice.

EvanMarkHopkins:
They do they don't even have to be bash, I would even opt for a build.is, or grunt etc

EvanMarkHopkins:
I just don't see the point of having a makefile here.

AdriVanHoudt:
yeah with more complex build you need something else but that's mainly a problem for front end website building

simon-p-r:
The point I think is it teaches participants another tool which is commonly used and I am all for  learning extras but it is not for everyone

AdriVanHoudt:
@simon-p-r true, but the makefile is almost 1-to-1 replaceable with npm in this case, the syntax is just different

simon-p-r:
@AdriVanHoudt true which means I am now learning both :+1:

EvanMarkHopkins:
Okay, but I'm still saying npm should be the runner of such weird scripts, having the makefile here creating barely shorter aliases to npm run <command> seems pointless

AdriVanHoudt:
@EvanMarkHopkins it's just an extra for people who prefer a Makefile aka Eran haha also @gyaresu sorry for taking over your PR for this discussion :p

EvanMarkHopkins:
@AdriVanHoudt is there a chat specific to this repo? I feel like General discussion pertaining to these assignments aren't entirely appropriate for Hapi's gitter either

AdriVanHoudt:
@EvanMarkHopkins I think general hapi gitter is fine and there is a gitter for every repo on github

gyaresu:
No. It's great. I'm really learning a lot.

I just found Evan's comments down the bottom of the Assignment 1 thread:  
https://github.com/hueniverse/hueniversity/issues/1

It's something I missed and there's some good info there.

For instance the reason for the global variable `internals`:
https://gist.github.com/hueniverse/a06f6315ea736ed1b46d

If we have a couple of days before the next assignment then the wiki would be worth giving some love:
https://github.com/hueniverse/hueniversity/wiki

EvanMarkHopkins:
Eran's*

AdriVanHoudt:
@gyaresu good point a lesson's learnt section would be helpful

EvanMarkHopkins:
Agree, also as a person not entirely familiar with github, how does a new participant go about catching up with previous assignments?

AdriVanHoudt:
@EvanMarkHopkins you can go to the closed issues for all the previous assignments https://github.com/hueniverse/hueniversity/issues?q=is%3Aissue+is%3Aclosed

EvanMarkHopkins:
Assuming everything after this one will include testing by default, that might be a huge turn off for someone who didn't have the opportunity to take part in this assignment

gyaresu:
@EvanMarkHopkins What he said :+1: 
So should we ask @hueniverse to enable Gitter? Or just pollute hapi/hapi on Gitter.

Also I agree about the hurdle for new participants.

That's why I think the README.md should point to the Wiki and we should index each Assignment.

We should also think about different levels of "hints"

So, Gitter or Freenode?

EvanMarkHopkins:
@gyaresu idk, but this is certainly something to start concidering

gyaresu:
@EvanMarkHopkins @AdriVanHoudt Is it bad form to open a new issue?

Project owner to decide on chat forum: Gitter.im or Freenode

Something like that? Mention wiki format for Assignments? Other things?

lib/index.js
```diff
@@ -1,28 +1,27 @@
...
-var Hoek = require('hoek');
 var Version = require('./version');
 
+var internals = {}; // Global variable https://gist.github.com/hueniverse/a06f6315ea736ed1b46d
 
-// Declare internals
-
-var internals = {};
-
+exports.init = function (port, callback) {
```
EvanMarkHopkins: Replace all instances of ```callback``` with ```next```, the reason behind such naming convensions are described by eran in #80 and in the [style guide](https://github.com/hapijs/contrib/blob/master/Style.md#callback)

EvanMarkHopkins:
@gyaresu It seems the wiki already does a pretty good job of documenting the 3rd assignment, I'd imagine it would be nice to have something similar for 1 and 2. I dont know @zoe-1 planned on doing this or not, but an issue may encourage someone to do it.

AdriVanHoudt:
@gyaresu you are fully free to open issues (even better to open to much than too little)

lib/start.js
```diff
@@ -0,0 +1,16 @@
+// Start.js - Why is this required over just index.js?
```
hueniverse: Because you don't want to expose the automatic server creation as an API.

gyaresu: Thank you.

PR #104, opened 2015-04-03
Assignment 3

anthonyringoet:


.eslintrc
```diff
@@ -0,0 +1,60 @@
+{
```
AdriVanHoudt: No need for this. Lab has a built in config file with Hapi rules (that you should enable on all tests with -L)

lib/index.js
```diff
@@ -1,28 +1,31 @@
...
 
-
 // Declare internals
-
+// https://gist.github.com/hueniverse/a06f6315ea736ed1b46d#module-classes
 var internals = {};
 
+exports.init = internals.init = function (port, next) {
+
+  if (typeof port === 'function') {
```
AdriVanHoudt: This requirement was removed from the assignment

lib/index.js
```diff
@@ -1,28 +1,31 @@
...
-        server.start(function (err) {
+    if (err){
+      return next(err, server);
+    }
 
-            Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
-        });
+    server.start(function startServer(err) {
+      return next(err, server);
```
AdriVanHoudt: newline after function

package.json
```diff
@@ -18,11 +18,16 @@
...
   "dependencies": {
+    "code": "^1.4.0",
     "hapi": "8.x.x",
-    "hoek": "2.x.x"
+    "hoek": "2.x.x",
+    "lab": "^5.5.1"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "make test-cov",
```
AdriVanHoudt: I prefer it the other way around (makefile --> npm) because windows doesn't have Make and like this you can't run it on windows but the other way around you can always run `npm test`

test/index.js
```diff
@@ -0,0 +1,69 @@
...
+      expect(err).to.be.equal('Error');
+      expect(server.info.port).to.equal(8000);
+      // set back
+      Version.register = backupRegister;
+
+      server.stop(done);
+    });
+
+  });
+
```
AdriVanHoudt: no newline required here

test/version.js
```diff
@@ -0,0 +1,36 @@
...
+var lab = exports.lab = Lab.script();
+var describe = lab.describe;
+var it = lab.it;
+var expect = Code.expect;
+
+describe('Version plugin/route', function () {
+
+  it('should return version on GET /version', function (done) {
+
+      Server.init(null, function initIndex(err, server) {
```
AdriVanHoudt: not really needed to name callback functions

test/version.js
```diff
@@ -0,0 +1,36 @@
...
+
+        expect(err, 'error').to.not.exist();
```
AdriVanHoudt: nice touch adding the prefix (also use it in your other tests ;))

PR #108, opened 2015-04-03
[Assignment 3] 100% Coverage

calesce:


.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: node_js
+node_js:
+  - 0.10
+  - 0.12
+  - iojs
```
EvanMarkHopkins: Look into configuring your editor, have it automatically insert newlines at the end of files. or test it yourself by running the program through `cat`

lib/index.js
```diff
@@ -10,19 +10,21 @@ var Version = require('./version');
 var internals = {};
 
 
-internals.init = function () {
+exports.init = function (port, callback) {
```
EvanMarkHopkins: replace all instances of callback with `next`, server.register using next internally, if you were using Hoek,nextTick, then you could name it w/e. check the style guide for more info, and a comment by Eran in #80

test/index.js
```diff
@@ -0,0 +1,42 @@
...
+
+            expect(err).to.equal(undefined);
+            expect(server.info.port).to.equal(8000);
+            server.stop(done);
+        });
+    });
+
+    it('starts the server with the given port', function (done) {
+
+        Server.init(3000, function (err, server) {
```
EvanMarkHopkins: lets use `0`, we want to avoid using specific ports while testing, and 0 will pick an available one for us. You can remove the check for port below, since not having an error is good enough for us.

lib/index.js
```diff
@@ -10,19 +10,21 @@ var Version = require('./version');
...
+
+    if (!callback) {
```
EvanMarkHopkins: if you're going the optional port route, lets check port's type here, and make sure it's a funciton. someone could just forget to put a callback, and then w/e port actually is will be assigned to callback

.travis.yml
calesce: @EvanMarkHopkins haha, I've never been a fan of the end of file newline, but it seems prevalent enough that I'll probably have to give in. thanks!

EvanMarkHopkins: yeah, it's part of the style guide, just being thorough :smile:

EvanMarkHopkins: that fact github marks it, is probably an indication it's probably for the best

PR #109, opened 2015-04-04
Assignment 3: 100% coverage

ipoddubny:
Ok, it should work by now.
However, I don't like the idea of mocking the Version module, it looks ugly and may break tests in parallel mode. Is there a better way to do it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,11 @@
     "hoek": "2.x.x"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "make test-cov",
     "start": "node lib/index.js"
+  },
+  "devDependencies": {
+    "code": "1.x.x",
+    "lab": "5.x.x"
   }
 }
--- a/test/index.js
+++ b/test/index.js
@@ -1,0 +1,57 @@
+// Load modules
+
+var Code = require('code');
+var Lab = require('lab');
+var Index = require('../lib/index');
+var Version = require('../lib/version');
+
+
+// Declare internals
+
+var internals = {};
+
+
+// Test shortcuts
+
+var lab = exports.lab = Lab.script();
+var describe = lab.experiment;
+var it = lab.test;
+var expect = Code.expect;
+
+
+describe('Index.init()', function () {
+
+    it('returns a started hapi server', function (done) {
+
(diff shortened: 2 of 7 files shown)


EvanMarkHopkins:
I think it's a reasonable solution given the constraints. We could configure additional plugins by passing init an options parameter, but only port is specified.

test/version.js
```diff
@@ -0,0 +1,43 @@
...
+var describe = lab.experiment;
+var it = lab.test;
+var expect = Code.expect;
+
+
+describe('Version.register()', function () {
+
+    it('registers /version route handler that returns version', function (done) {
+
+        var server = new Hapi.Server();
```
EvanMarkHopkins: I think we can safely use the init for testing version, but you could look into creating a new http server and passing the failing plugin, that would a pretty decent solution in my opinion

test/version.js
```diff
@@ -0,0 +1,43 @@
...
+
+        server.register(Version, function (err) {
+
+            expect(err).to.not.exist();
+
+            server.inject('/version', function (res) {
+
+                expect(res.statusCode).to.equal(200);
+                expect(res.result).to.contain('version');
+                expect(res.result.version).to.equal('0.0.2');
```
EvanMarkHopkins: Let's use the actual package.version for comparison, we shouldn't have to update this test every time the version changes. Everything else looks good :+1:

package.json
```diff
@@ -22,7 +22,11 @@
     "hoek": "2.x.x"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "make test-cov",
```
AdriVanHoudt: this breaks for windows users

test/version.js
AdriVanHoudt: maybe also test for the exact result. If the endpoint returns an object with an extra key the test will never catch it

test/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+    it('returns a started hapi server', function (done) {
+
+        Index.init(null, function (err, server) {
+
+            expect(err).to.not.exist();
+
+            expect(server.info.started).to.be.above(0);
+
+            server.stop(done);
```
hussainanjar: Why space after each line ?

PR #112, opened 2015-04-12
Assignment3

### LINKED ISSUE #79: 100% coverage

tielur:
Completed Assignment 3

Closes #79


lib/start.js
```diff
@@ -0,0 +1,10 @@
+var Index = require('./index');
+
+Index.init(8000, function (error, server) {
+
+  if (error) {
+      console.log(error);
```
AdriVanHoudt: use `Hoek.assert(!error, error)`, it will automatically throw when an error is present and you don't need the if else then

package.json
```diff
@@ -22,7 +22,12 @@
     "hoek": "2.x.x"
   },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "start": "node lib/index.js"
+    "test": "make test-cov",
```
AdriVanHoudt: I prefer the makefile pointing to npm because this won't work on windows

test/index.js
```diff
@@ -0,0 +1,43 @@
...
+var describe = lab.describe;
+var it = lab.it;
+var before = lab.before;
+var after = lab.after;
+var expect = Code.expect;
+
+describe('Index', function () {
+
+    it('successfully starts', function(done) {
+        Index.init(0, function(err, server) {
```
AdriVanHoudt: just pass null as port, Hapi will resolve to 0 automaticly

package.json
tielur: I pair programmed with a friend on this, and he uses windows. We got `make` to work by using this:
http://gnuwin32.sourceforge.net/packages/make.htm

It's already a pain to try to do development with node on windows as it is, so I can see your point on not making it any harder though.

AdriVanHoudt: yeah I also made it work but why the extra trouble :p

lib/start.js
```diff
@@ -0,0 +1,8 @@
+var Hoek = require('hoek');
+var Index = require('./index');
+
+Index.init(8000, function (error, server) {
```
AdriVanHoudt: check http://hapijs.com/styleguide#user-content-variable-names for naming error vars

test/index.js
```diff
@@ -0,0 +1,43 @@
...
+var describe = lab.describe;
+var it = lab.it;
+var before = lab.before;
+var after = lab.after;
+var expect = Code.expect;
+
+describe('Index', function () {
+
+    it('successfully starts', function(done) {
+        Index.init(null, function(err, server) {
```
AdriVanHoudt: blank line at the beginning of function scope (aka newline after `function() {`)

tielur: @AdriVanHoudt is this a common style used on Hapi projects? If so, I wonder if eslint would accept a rule like `newline-after-function` similar to the existing [`newline-after-var`](http://eslint.org/docs/rules/newline-after-var)

AdriVanHoudt: @tielur yes it is and @cjihrig beat you to it ;) https://github.com/hapijs/lab/pull/339

test/index.js
```diff
@@ -0,0 +1,44 @@
...
+describe('Index', function () {
+
+    it('successfully starts', function(done) {
+
+        Index.init(null, function(err, server) {
+
+            expect(err).to.not.exist();
+            server.stop(done);
+        });
+
```
AdriVanHoudt: no need for a newline here

test/version.js
```diff
@@ -0,0 +1,58 @@
...
+
+var describe = lab.describe;
+var it = lab.it;
+var before = lab.before;
+var after = lab.after;
+var expect = Code.expect;
+
+describe('Version Plugin', function () {
+
+    var server = new Hapi.Server();
```
AdriVanHoudt: if you want to do this better to use `before` but in this case it is better to use a new server for every test so if 1 fails not all of them fail

test/version.js
```diff
@@ -0,0 +1,58 @@
...
+
+        server.register(Version, function(err) {
+
+            expect(err).to.not.exist();
+            done();
+        });
+
+    });
+
+    it('registers routes', function(done) {
```
AdriVanHoudt: this is also tested in the test below so not really needed

test/version.js
```diff
@@ -0,0 +1,58 @@
...
+            method: 'GET',
+            url: '/version'
+        };
+
+        server.inject(options, function(response) {
+          
+            var result = response.result;
+
+            expect(response.statusCode).to.equal(200);
+            expect(result).to.be.instanceof(Object);
```
AdriVanHoudt: easier and better, since you test for the exact object, to just `expect(response.result).to.deep.equal({version: Package.version})`

test/version.js
tielur: Yea I'm used to the before block coming from rspec, but this assignment said to avoid those for now. I'll create a new server per test for now.

test/version.js
tielur: ahh I like that a lot!

test/version.js
tielur: Yea I've been wondering about this actually, like how you would test explicitly if adding a route fails. There seems to be quite a few plugins that do things like add routes or hook into the server extensions but without a real way to report back if it failed or not. I suppose just testing that those work like you mentioned tests that they registered directly.

test/version.js
AdriVanHoudt: :+1:

tielur: pushed a change, the only down side is that I have to register the plugin in each test now.

test/version.js
```diff
@@ -0,0 +1,52 @@
...
+                method: 'GET',
+                url: '/version'
+            };
+
+            server.inject(options, function(response) {
+
+                var result = response.result;
+
+                expect(response.statusCode).to.equal(200);
+                expect(response.result).to.deep.equal({version: Package.version});
```
AdriVanHoudt: space after and before {} when doing inline object notation

.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: node_js
+node_js:
+  - "0.12"
+  - "0.10"
+  - "iojs"
```
tielur: Should I be adding a new line here?

AdriVanHoudt: not needed per se but you can if you will for cleanness ^^

tielur: Changed it but apparently Github didn't notice it to make this outdated.

AdriVanHoudt: maybe not special enough for github :disappointed:

PR #113, opened 2015-04-12
assignment3

haikyuu:
*  lib/index.js now exports init function
  * Add travis.yml
  * introduce start.js as new entry point
  * introduce lab & code for testing
  * 100% test coverage

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,12 +8,12 @@ var Version = require('./version');
 var internals = {};
 
 
-exports.init = internals.init = function (port, callback) {
+exports.init = internals.init = function (port, next) {
 
     var defaultPort = 8000;
     //support both init(port, cb) & init(cb)
     if ( typeof port === 'function' ) {
-        callback = port;
+        next = port;
         port = defaultPort;
     }
     if ( port === undefined || port === null ) {
@@ -24,11 +24,11 @@ exports.init = internals.init = function (port, callback) {
     server.register(Version, function (err) {
 
         if (err) {
-            return callback(err);
+            return next(err);
         }
         server.start(function (err) {
 
-            return callback(err, server);
+            return next(err, server);
         });
     });
 };
--- a/lib/index.js
+++ b/lib/index.js
@@ -7,22 +7,24 @@ var Version = require('./version');
 
 // Declare internals
 
-var internals = {};
+var internals = {
+    errors: []
+};
 
 
-internals.init = function () {
+exports.init = internals.init = function (port, callback) {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port });
     server.register(Version, function (err) {
 
         Hoek.assert(!err, err);
+        internals.errors.push(err);
         server.start(function (err) {
 
             Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+            internals.errors.push(err);
+            callback(internals.errors, server);
         });
     });
 };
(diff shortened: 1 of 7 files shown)


lib/index.js
```diff
@@ -7,22 +7,24 @@ var Version = require('./version');
...
-internals.init = function () {
+exports.init = internals.init = function (port, callback) {
 
     var server = new Hapi.Server();
-    server.connection({ port: 8000 });
+    server.connection({ port: port });
     server.register(Version, function (err) {
 
         Hoek.assert(!err, err);
+        internals.errors.push(err);
```
AdriVanHoudt: you will never push something here since when there is an error `Hoek.assert()` will throw

lib/index.js
```diff
@@ -7,22 +7,24 @@ var Version = require('./version');
...
         server.start(function (err) {
 
             Hoek.assert(!err, err);
-            console.log('Server started at: ' + server.info.uri);
+            internals.errors.push(err);
+            callback(internals.errors, server);
```
AdriVanHoudt: use `return callback()` for readability

package.json
```diff
@@ -21,8 +21,12 @@
...
     "hoek": "2.x.x"
   },
+  "devDependencies": {
+    "code": "^1.4.0",
+    "lab": "^5.5.1"
+  },
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "start": "node lib/index.js"
+    "test": "make test-cov",
```
AdriVanHoudt: As I've stated before, I prefer Makefile pointing to npm since this won't work on windows

test/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+
+describe('server initiaion', function () {
+
+    it('initiate a server without errors', function (done) {
+		Init(8000, function (errors, server) {
+
+			expect(errors).to.only.include(undefined);
+			server.stop();
+			done();
```
AdriVanHoudt: do `server.stop(done)` so you are sure to only stop the test when the server is fully stopped

haikyuu:
Thanks @AdriVanHoudt  .
I borrowed your test which alters the register function of the plugin, nice trick (i totally forgot that the required modules were cached).

haikyuu:
Wouldn't it be better if we required all the plugins in the lib folder and registered them **once**.
We could use [require-dir] (https://github.com/aseemk/requireDir) or something like
```javascript
Fs.readdirSync(__dirname).filter(function(file) {

        return ((file !== 'index.js') && (file.indexOf('.js') !== -1);
    }).forEach(function(file) {
        internals.plugins.push(require(Path.join(__dirname, file)));
    });
```

AdriVanHoudt:
@haikyuu np and we register them once right now...

haikyuu:
@AdriVanHoudt  i meant that we shouldn't have to update the index.js file every time we add a plugin.

AdriVanHoudt:
@haikyuu seems logical to me, since every plugin has specific options, we can however in the future use https://github.com/hapijs/glue

test/index.js
```diff
@@ -0,0 +1,94 @@
...
+
+		var port = 8002;
+		Init(port, function (err, server) {
+
+			expect(err, 'err').not.to.exist();
+			expect(server.info.port, 'port').to.equal(port);
+			server.stop(done);
+		});
+
+    });
```
hueniverse: Need to work on style guide rules. New lines and spaces are wrong.

PR #117, opened 2015-04-19
set parallel to false when monkey patching

### LINKED ISSUE #116: Parallel test execution
Shouldn't the parallel option set [here](https://github.com/hueniverse/hueniversity/blob/master/test/index.js#L41) be false? Lab's default for parallel execution seems to be false but I think the point was to explicitly mark this one as basically not safe for parallel execution if we change the default for the tests to parallel in the future.

As noted [here](https://github.com/hueniverse/hueniversity/issues/79#issuecomment-93561251):
> When monkey patching code in a test, mark that test with the { parallel: false } lab option to make it both safe for future parallel testing as well as visual cue.

tielur:
Fixes #116

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -38,7 +38,7 @@ describe('Server', function() {
         });
     });
 
-    it('handles register plugin errors', { parallel: true }, function (done) {
+    it('handles register plugin errors', { parallel: false }, function (done) {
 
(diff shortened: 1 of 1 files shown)


hueniverse:
Didn't notice this one.

PR #119, opened 2015-04-20
Assingment 4

### LINKED ISSUE #118: Basic authentication

Due: 4/27

AdriVanHoudt:
closes #118 

* I added the `v` option to test command because it's easier to see what is going wrong with your tests and since this is a learning thing I thought it might help (it helped me)
* Removed `describe` shortcut in test/index.js
* Renamed `Server` to `Hueniversity` test/version.js
    *  see https://github.com/hueniverse/hueniversity/commit/13b9f07922ab36fe4dde0479638d79baf6f099e5 for more explanation
* Added linting thresholds to make the tests fail when linting returns errors or warnings

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 test:
 	@node node_modules/lab/bin/lab -a code
 test-cov:
-	@node node_modules/lab/bin/lab -a code -t 100 -L
+	@node node_modules/lab/bin/lab -a code -t 100 -Lv
 test-cov-html:
 	@node node_modules/lab/bin/lab -a code -r html -o coverage.html
 
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 test:
 	@node node_modules/lab/bin/lab -a code
 test-cov:
-	@node node_modules/lab/bin/lab -a code -t 100 -Lv
+	@node node_modules/lab/bin/lab -a code -t 100 -Lv --lint-errors-threshold 0 --lint-warnings-threshold 0
 test-cov-html:
 	@node node_modules/lab/bin/lab -a code -r html -o coverage.html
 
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
 test:
 	@node node_modules/lab/bin/lab -a code
 test-cov:
-	@node node_modules/lab/bin/lab -a code -t 100 -L
+	@node node_modules/lab/bin/lab -a code -t 100 -Lv --lint-errors-threshold 0 --lint-warnings-threshold 0
```
hueniverse: Are these extra flags really needed?

AdriVanHoudt: Debatable but Imo they will help prevent that the majority of the comments
on pr's are about following the hapi style.

lib/users.json
```diff
@@ -0,0 +1,10 @@
+{
+	"Foo": {
+		"username": "Foo",
```
hueniverse: Safer to add the username programatically than repeat information.

test/index.js
```diff
@@ -50,7 +49,7 @@ it('handles register plugin errors', { parallel: false }, function (done) {
         name: 'fake version'
     };
 
-    Hueniversity.init(0, function (err, server) {
```
hueniverse: Better to keep the callback function with all its actual arguments. Kind of self documenting.

lib/users.json
AdriVanHoudt: you mean not have the username and assume the key is?

test/index.js
AdriVanHoudt: Seems like a waste to pass around unused objects, the callback from the init function is self documenting enough imo, but you have a point

PR #120, opened 2015-04-20
Assignment #4 - Basic authentication

### LINKED ISSUE #118: Basic authentication

hussainanjar:
Closes #118

### DIFF
--- a/lib/private.js
+++ b/lib/private.js
@@ -41,9 +41,9 @@ exports.register = function(server, options, next) {
             }
         });
 
-    });
+        return next();
 
-    return next();
+    });
 };
 
 exports.register.attributes = {
--- a/lib/private.js
+++ b/lib/private.js
@@ -1,0 +1,51 @@
+// Load modules
+
+var HapiAuthBasic = require('hapi-auth-basic');
+var Users = require('./users.json');
+
+// Declare internals
+
+var internals = {};
+
+internals.validate = function (username, password, callback) {
+
+    var user = Users[username];
+    if (!user || password !== user.password) {
+        return callback(null, false);
+    }
+
+    return callback(null, true, { user: user });
+};
+
+exports.register = function(server, options, next) {
+
+    server.register(HapiAuthBasic, function (err) {
+
+        if (err) {
+            return next(err);
+        }
+
+        server.auth.strategy('simple', 'basic', { validateFunc: internals.validate });
+
+        server.route({
+            method: 'GET',
+            path: '/private',
+            config: {
+                description: 'This is a private endpoint',
+                auth: 'simple',
+                handler: function(request, reply) {
+
+                    var html = 'Welcome ' + request.auth.credentials.user.username;
+                    return reply(html);
+                }
+            }
+        });
+
+    });
+
+    return next();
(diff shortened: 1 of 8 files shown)


FennNaten:
Hi,
I can see that you have the same failing test issue that I had. I don't want to spoil the fun of bug hunting, so I won't write the solution here, but if you're stuck and want help at some point, you can ask ;)
Also, sorry for the reference to your Pull Request that I made, I wanted to link the issue but failed to select the correct item in the suggestions -_-

hussainanjar:
@FennNaten  help :smile_cat:

FennNaten:
^^ Just look at my latest commit on pr #121 ;) if you copy/pasted version.js as I did for private.js base, your next() call is not in the right place. You should move it into the callback of register(basic), otherwise it's called too soon ;)

lib/private.js
```diff
@@ -0,0 +1,51 @@
...
+var internals = {};
+
+internals.validate = function (username, password, callback) {
+
+    var user = Users[username];
+    if (!user || password !== user.password) {
+        return callback(null, false);
+    }
+
+    return callback(null, true, { user: user });
```
AdriVanHoudt: why not just pass the user object? Now you'll have to access it via `request.auth.credentials.user` instead of `request.auth.credentials`

lib/version.js
```diff
@@ -2,7 +2,6 @@
 
 var Pkg = require('../package.json');
 
-
```
AdriVanHoudt: why remove these lines? They're according to the style guide

package.json
```diff
@@ -19,6 +19,7 @@
   "homepage": "https://github.com/hueniverse/hueniversity",
   "dependencies": {
     "hapi": "8.x.x",
+    "hapi-auth-basic": "^2.0.0",
```
AdriVanHoudt: try using the 2.x.x notation, reads better

test/private.js
```diff
@@ -0,0 +1,112 @@
...
+var describe = lab.experiment;
+var expect = Code.expect;
+var it = lab.test;
+
+
+describe('/private', function () {
+
+    it('should ask for authentication', function (done) {
+
+        Server.init(0, function (err, server) {
```
AdriVanHoudt: see https://github.com/hueniverse/hueniversity/commit/13b9f07922ab36fe4dde0479638d79baf6f099e5#commitcomment-10797003

lib/private.js
hussainanjar: Naturally it feels like like passing object is more useful, for instance if the user object had firstName and lastName or roles, i would get them with the object.

package.json
hussainanjar: Yes it does but this got added using command `npm install --save hapi-auth-basic` I wouldn't want to change the default npm behaviour every time or am I doing something wrong ?

AdriVanHoudt: yeah I know it's a pain :/

lib/private.js
hueniverse: I agree with @AdriVanHoudt. The user key doesn't add much here.

PR #121, opened 2015-04-20
Assignment 4: add basic auth

FennNaten:
hueniversity assignment 4 (solves #118 )

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,7 +2,7 @@
 
 var Hapi = require('hapi');
 var Version = require('./version');
-
+var Private = require('./private');
 
 // Declare internals
 
@@ -14,7 +14,7 @@ exports.init = function (port, next) {
     var server = new Hapi.Server();
     server.connection({ port: port });
 
-    server.register(Version, function (err) {
+    server.register([Version, Private], function (err) {
 
         if (err) {
             return next(err);
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,12 +9,16 @@ var Private = require('./private');
 var internals = {};
 
 
-exports.init = function (port, next) {
+exports.init = function (config, next) {
+
+    if (!config || typeof (config) !== 'object'){
+        return next(new Error('mandatory configuration object not provided'));
+    }
 
     var server = new Hapi.Server();
-    server.connection({ port: port });
+    server.connection({ port: config.port });
 
-    server.register([Version, Private], function (err) {
+    server.register([Version, {register: Private, options: {users: config.users}}], function (err) {
 
         if (err) {
             return next(err);
--- a/lib/private.js
+++ b/lib/private.js
@@ -1,0 +1,59 @@
+// Load modules
+
+var Users = require('./users.json');
+var Basic = require('hapi-auth-basic');
+
+
+// Declare internals
+
+var internals = {
+    response: {
+        template: '<html><head><title>private page</title></head><body>Greetings %username%. Welcome to the private section.</body></html>',
+        generate: function (username){
+
+            return this.template.replace('%username%', username);
+        }
+    },
+    authStrategyName: 'private-basic',
+    authValidate: function(username, password, callback) {
+
+        var user = Users[username];
+        if (!user) {
+            return callback(null, false);
+        }
+        if (user.password === password){
+            return callback(null, true, {username: user.username});
+        }
+        callback(null, false);
+    }
+};
+
+
+exports.register = function (server, options, next) {
+
+    server.register(Basic, function(err){
+        if (err) {
+            return next(err);
+        }
+        server.auth.strategy(internals.authStrategyName, 'basic', { validateFunc: internals.authValidate });
+
+        server.route({
+            method: 'GET',
+            path: '/private',
+            config: {
+                auth: internals.authStrategyName,
+                description: 'Returns a welcome message if user successfully authenticated via basic auth',
+                handler: function (request, reply) {
+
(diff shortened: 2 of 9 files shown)


lib/private.js
```diff
@@ -0,0 +1,60 @@
...
+var internals = {
+    response: {
+        template: '<html><head><title>private page</title></head><body>Greetings %username%. Welcome to the private section.</body></html>',
+        generate: function (username){
+
+            return this.template.replace('%username%', username);
+        }
+    },
+    authStrategyName: 'private-basic',
+    authValidate: function(username, password, callback) {
```
AdriVanHoudt: personally I would declare the function outside of the object declaration as in `internals.authValidate = funtion()`

lib/private.js
```diff
@@ -0,0 +1,60 @@
...
+
+        var user = Users[username];
+        if (!user) {
+            return callback(null, false);
+        }
+        if (user.password === password){
+            return callback(null, true, {username: user.username});
+        }
+        callback(null, false);
```
AdriVanHoudt: why not return here?

lib/private.js
```diff
@@ -0,0 +1,60 @@
...
+                handler: function (request, reply) {
+
+                    return reply(internals.response.generate(request.auth.credentials.username));
+                }
+            }
+        });
+
+        return next();
+    });
+
```
AdriVanHoudt: whiteline not needed

lib/private.js
FennNaten: Good question. As I understood the behavior of the validation function, its return value is never used, and it entirely relies on the callback for validation result.  
So I basically put return keywords only to stop method execution when invoking the callback, which wasn't needed for the last one as end of scope was reached already.  
Now that you mention it, it may damage readability and fail to convey intent.  
It may be better to strip out the return statements and replace the code with a if/else if/else structure.

lib/private.js
FennNaten: hum, that's what I did in the tests of the file, mainly because I copied their overall style from hapi-auth-basic plugin tests. 
Naturally, I tend to define all of an object properties on the spot when their existence is unconditional. Could you please elaborate on why you would declare it outside? I don't see the benefits but I may miss something.

lib/private.js
FennNaten: true. I haven't yet assimilated the style guide ^^'

lib/private.js
AdriVanHoudt: no the returns are good, basic rule is to put where you, you never guess, return. And since the callback is (most of time and in this case) the place where you return it helps readability to also put a return before this callback

lib/private.js
AdriVanHoudt: this is a minor one so np, also there are talks(and work) about improving the linter in lab to report more on the required styleguide so it will only get easier :D

lib/private.js
FennNaten: roger that! ;)

lib/private.js
AdriVanHoudt: the only benefit is readability. If the function gets large or you have a lot of them it can get messy but of course this is more of a personal note than an official style from Hapi

FennNaten:
I wasn't satisfied by directly using 'production' users in the tests, so I modified Users structure and tests to inject users before each test and switch back to original list after. 
However, this still doesn't feel right. I'll give a shot at more refactoring to have Users module injected in the private plugin at server instantiation, so that mocking is made easier. 
I committed anyway, because this is a learning experiment, so I wanted the PR to reflect thought process. ;)

AdriVanHoudt:
The users are just a placeholder right now, see the assignment, this will later probably be stored in a db which allows you to make a test db.
If you really want to do this you might be better of with having a users file for testing and require the right one base on `process.env.NODE_ENV` which standards to `test` when running Lab

FennNaten:
Yes, I know that's only a placeholder and the code will be further refactored later, if I was just prototyping and chaining tasks I think I would not care as much, and wait for the next round. However I have some time, so I use it to do some experiments ;)
I'm generally not a fan of swapping logic based on environment, except for a config file maybe, if this file's keys are used to feed explicit initialization of elements when an element is bootstrapped.
I know that feels a bit like over engineering at this stage ;)

AdriVanHoudt:
You can consider the users files config files ;)

FennNaten:
Here a version that feels better (to me), where users module is injected during application bootstrapping phase instead of being required in private module, so that fake users can be injected in tests. 
Would still be best in my opinion to have some config files for the app, but I'll wait and see how things evolve on next assignments. ;)

AdriVanHoudt:
this only has a point if you at least pass in 'test users' instead of the same users you would normally use :D

FennNaten:
Indeed. The previous commit also took care of that ;)

lib/index.js
```diff
@@ -2,19 +2,23 @@
...
-
+var Private = require('./private');
 
 // Declare internals
 
 var internals = {};
 
 
-exports.init = function (port, next) {
+exports.init = function (config, next) {
```
hueniverse: passing the user file via config wasn't in the assignment.

lib/private.js
```diff
@@ -0,0 +1,66 @@
+// Load modules
+
+var Basic = require('hapi-auth-basic');
+
+// Declare internals
+
+var internals = {
+    response: {
+        template: '<html><head><title>private page</title></head><body>Greetings %username%. Welcome to the private section.</body></html>',
```
hueniverse: No templates was in the assignment.

PR #124, opened 2015-04-23
Assignment 4 basic Auth

Valuuin:
closes Assignment 4 #118

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -1,0 +1,53 @@
+var Auth = require('hapi-auth-basic');
+var Users = require('./users.json');
+
+// Declare internals
+
+var internals = {};
+
+internals.validate = function (username, password, callback) {
+
+    var user = Users[username];
+
+    if (!user){
+        return callback(null, false);
+    }
+
+    if ( user.password === password ){
+        callback(null, true, { username: user.name });
+    }
+    else {
+        callback(null, false);
+    }
+
+};
+
+exports.register = function (server, options, next) {
+
+    server.register(Auth, function (err){
+
+        server.auth.strategy('simple', 'basic', { validateFunc: internals.validate } );
+
+        server.route({
+            method: 'GET',
+            path: '/private',
+            config: {
+
+                auth: 'simple',
+                description: 'Returns the version of the server',
+                handler: function (request, reply) {
+
+                    return reply('<p> Welcome, ' + request.auth.credentials.username + '</p>');
+                }
+            }
+        });
+
+        return next();
+
+    });
+
+};
+
+exports.register.attributes = {
+    name: 'authentication'
+};
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,7 +2,7 @@
 
 var Hapi = require('hapi');
 var Version = require('./version');
-
+var Auth = require('./auth');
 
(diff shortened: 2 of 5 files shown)


lib/auth.js
```diff
@@ -0,0 +1,53 @@
...
+
+    if (!user){
+        return callback(null, false);
+    }
+
+    if ( user.password === password ){
+        callback(null, true, { username: user.name });
+    }
+    else {
+        callback(null, false);
```
AdriVanHoudt: why not try to combine these if's since you are returning (null, false) twice

lib/auth.js
```diff
@@ -0,0 +1,53 @@
...
+
+    server.register(Auth, function (err){
+
+        server.auth.strategy('simple', 'basic', { validateFunc: internals.validate } );
+
+        server.route({
+            method: 'GET',
+            path: '/private',
+            config: {
+
```
AdriVanHoudt: no need for a newline here

lib/auth.js
```diff
@@ -0,0 +1,53 @@
...
+                description: 'Returns the version of the server',
+                handler: function (request, reply) {
+
+                    return reply('<p> Welcome, ' + request.auth.credentials.username + '</p>');
+                }
+            }
+        });
+
+        return next();
+
```
AdriVanHoudt: also no newline here

lib/index.js
```diff
@@ -25,4 +25,5 @@ exports.init = function (port, next) {
             return next(err, server);
         });
     });
+
```
AdriVanHoudt: also no newline here :p

package.json
```diff
@@ -19,6 +19,7 @@
   "homepage": "https://github.com/hueniverse/hueniversity",
   "dependencies": {
     "hapi": "8.x.x",
+    "hapi-auth-basic": "^2.0.0",
```
AdriVanHoudt: try using the 8.x.x notation

test/auth.js
```diff
@@ -0,0 +1,82 @@
...
+        Server.init(0, function (err, server) {
+
+            expect(err).to.not.exist();
+
+            var request = { url: '/private', headers: { authorization: internals.header('john', '12345') } };
+
+            server.inject(request, function (res) {
+
+                expect(res.statusCode).to.equal(200);
+                expect(res.result).to.include('John Doe');
```
AdriVanHoudt: why not match the whole object to make the test more fool proof?

lib/auth.js
```diff
@@ -0,0 +1,53 @@
...
+    else {
+        callback(null, false);
+    }
+
+};
+
+exports.register = function (server, options, next) {
+
+    server.register(Auth, function (err){
+
```
AdriVanHoudt: you are not catching the error here

lib/auth.js
```diff
@@ -0,0 +1,53 @@
+var Auth = require('hapi-auth-basic');
```
FennNaten: In this file you're naming 'Auth' the hapi-auth-basic plugin, and in index.js file 'Auth' is instead auth.js. I think you should use different names across the app for this two things, to avoid confusion.

PR #126, opened 2015-04-23
Assignment 4 - Authentication

### LINKED ISSUE #118: Basic authentication

tielur:
closes #118

### DIFF
--- a/lib/private.js
+++ b/lib/private.js
@@ -1,0 +1,53 @@
+// Load modules
+
+var HapiAuthBasic = require('hapi-auth-basic');
+var Users = require('./users.json');
+
+
+// Declare internals
+
+var internals = {};
+
+internals.validate = function (username, password, callback) {
+
+  var user = Users[username];
+  if (!user) {
+      return callback(null, false);
+  }
+
+  var isValid = password === user.password;
+
+  return callback(null, isValid, { username: user.username});
+};
+
+
+exports.register = function (server, options, next) {
+
+  server.register(HapiAuthBasic, function (err) {
+
+    if (err) {
+      return next(err);
+    }
+
+    server.auth.strategy('simple', 'basic', { validateFunc: internals.validate });
+
+    server.route({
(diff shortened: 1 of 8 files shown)


lib/private.js
```diff
@@ -0,0 +1,53 @@
...
+  return callback(null, isValid, { username: user.username});
+};
+
+
+exports.register = function (server, options, next) {
+
+  server.register(HapiAuthBasic, function (err) {
+
+    if (err) {
+      return next(err);
```
hussainanjar: Your indentation is not 4 spaces consistently.

tielur: Thanks @hueniverse. I ended up copying the .eslintrc and .eslintignore from lab and putting it in the root directory of this project. I then turned on indentationn to match the style we are trying to follow:

```
"indent": [2, 4]
```

http://eslint.org/docs/rules/indent.html


Is it common to do this with Lab? Lab has a default for both of these files, but to add a new rule or change rules, you just copy it and make modifications? Should these be committed to the project?

tielur:
I opened this pull request over at lab to hopefully catch things like this in the future.

https://github.com/hapijs/lab/pull/349

hussainanjar:
@tielur :+1:

tielur:
Lab was updated, so you should now get a warning for indentation.  Something like this:

```
Linting results:
	lib/index.js:
		Line 15: indent - Expected indentation of 4 characters.
```

FennNaten:
Nice!

lib/private.js
```diff
@@ -0,0 +1,53 @@
...
+internals.validate = function (username, password, callback) {
+
+    var user = Users[username];
+    if (!user) {
+        return callback(null, false);
+    }
+
+    var isValid = password === user.password;
+
+    return callback(null, isValid, { username: user.username});
```
idanwe: missing before `}`

PR #127, opened 2015-04-25
Assignment 4

TheAlphaNerd:
* Adds hapi-basic-auth 
* Adds route ```/private```
* Adds stubbed ```lib/users.json``` to emulate login\8
* 100% test coverage

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,6 +2,7 @@
 
 var Hapi = require('hapi');
 var Version = require('./version');
+var Private = require('./private');
 
 
 // Declare internals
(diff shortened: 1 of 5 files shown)


lib/index.js
```diff
@@ -14,7 +15,7 @@ exports.init = function (port, next) {
     var server = new Hapi.Server();
     server.connection({ port: port });
 
-    server.register(Version, function (err) {
+    server.register([require('hapi-auth-basic'), Private, Version], function (err) {
```
hueniverse: Better is the Private plugin takes care of it's own dependencies (e.g. the basic auth plugin).

PR #128, opened 2015-04-28
trying assignment 4

mikejerome:
I was too busy in the last week and ran out of time for this one :cry:

### DIFF
--- a/lib/private.js
+++ b/lib/private.js
@@ -1,0 +1,48 @@
+// Load modules
+var Basic = require('hapi-auth-basic');
+var Users = require('./users');
+
+
+// Declare internals
+
+var internals = {};
+
+internals.validate = function (username, password, callback) {
+
+    var user = Users[username];
+    if (!user || user.password !== password) {
+        return callback(null, false);
+    }
+
+    return callback(null, true, user);
+};
+
+exports.register = function (server, options, next) {
+
+    server.register(Basic, function (err) {
+
+        if (err) {
+            return next(err);
+        }
+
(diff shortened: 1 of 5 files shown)


lib/private.js
```diff
@@ -0,0 +1,48 @@
...
+        server.auth.strategy('simple', 'basic', { validateFunc: internals.validate });
+        server.route({
+            method: 'GET',
+            path: '/private',
+            config: {
+                description: 'Returns welcome page protected by basic auth',
+                auth: 'simple',
+                handler: function (request, reply) {
+
+                    return reply('hello');
```
AdriVanHoudt: the reply should at least have the (user)name of the logged in user in it

package.json
```diff
@@ -19,6 +19,7 @@
   "homepage": "https://github.com/hueniverse/hueniversity",
   "dependencies": {
     "hapi": "8.x.x",
+    "hapi-auth-basic": "^2.0.0",
```
AdriVanHoudt: try using the 2.x.x style notation

test/private.js
```diff
@@ -0,0 +1,98 @@
...
+                expect(res.statusCode, 'Status code').to.equal(401);
+
+                server.stop(done);
+            });
+        });
+    });
+
+    it('returns error on wrong username', function (done) {
+
+        Hueniversity.init(0, function (err) {
```
idanwe: you are missing the server argument `(err, server)`

mikejerome:
Thanks a lot for the help @idanwe and @AdriVanHoudt

PR #129, opened 2015-04-28
Assignment 4

### LINKED ISSUE #118: Basic authentication

idanwe:
* Add /private route
 * Add hapi-auth-basic
 * Rename Server to Huenversity
 * Closes #118

### DIFF
--- a/lib/private.js
+++ b/lib/private.js
@@ -1,0 +1,56 @@
+// Load modules
+
+var Basic = require('hapi-auth-basic');
+var UsersData = require('./users.json');
+
+// Declare internals
+
+var internals = {};
+
+
+exports.register = function (server, options, next) {
+
+    server.register(Basic, function (err) {
+
+        if (err) {
+            return next(err);
+        }
+
+        server.auth.strategy('simple', 'basic', { validateFunc: internals.validate });
+
+        server.route({
+            method: 'GET',
+            path: '/private',
+            config: {
+                description: 'Returns welcome message with the authenticated username',
+                auth: 'simple',
+                handler: function (request, reply) {
+
+                    var username = request.auth.credentials.username;
+                    return reply(internals.response(username));
+                }
+            }
+        });
+
+        return next();
+    });
+};
+
+exports.register.attributes = {
+    name: 'private'
+};
+
+internals.validate = function (username, password, callback) {
+
+    var user = UsersData[username];
+    if (user && user.password === password) {
+        return callback(null, true, user);
+    }
+
+    return callback(null, false);
+};
+
+internals.response = function (username) {
+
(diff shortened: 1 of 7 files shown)


lib/private.js
```diff
@@ -0,0 +1,57 @@
...
+
+                    var username = request.auth.credentials.username;
+                    return reply(internals.respone(username));
+                }
+            }
+        });
+
+        return next();
+    });
+
```
AdriVanHoudt: no need for newline here

lib/private.js
```diff
@@ -0,0 +1,57 @@
...
+};
+
+exports.register.attributes = {
+    name: 'private'
+};
+
+internals.validate = function (username, password, callback) {
```
AdriVanHoudt: personally I prefer declaring functions before usage

test/private.js
```diff
@@ -0,0 +1,114 @@
...
+            var request = {
+                url: '/private',
+                headers: {
+                    authorization: internals.header('user1', 'secret1')
+                }
+            };
+            server.inject(request, function (res) {
+
+                expect(res.statusCode).to.equal(200);
+                expect(res.result).to.contains('user1');
```
AdriVanHoudt: matching the full res.result object would tighten up the test, imagine the server adding an extra key, this test will not catch that

lib/private.js
idanwe: I think it is ok, for example from the [hoek code](https://github.com/hapijs/hoek/blob/master/lib/index.js) 

And I find it more clean

AdriVanHoudt: yeah true, probably because of my Java background haha

FennNaten:
Hi,
I can see that you didn't updated the package.json to add hapi-auth-basic module. Maybe you forgot to install it with --save flag.

lib/private.js
```diff
@@ -0,0 +1,56 @@
...
+        server.route({
+            method: 'GET',
+            path: '/private',
+            config: {
+                description: 'Returns welcome message with the authenticated username',
+                auth: 'simple',
+                handler: function (request, reply) {
+
+                    var username = request.auth.credentials.username;
+                    return reply(internals.respone(username));
```
FennNaten: little typo here, you typed 'respone' instead of 'response'

lib/private.js
```diff
@@ -0,0 +1,56 @@
...
+
+    var user = UsersData[username];
+    if (user && user.password === password) {
+        return callback(null, true, user);
+    }
+
+    return callback(null, false);
+};
+
+internals.respone = function (username) {
```
FennNaten: same typo as above ;)

idanwe: @FennNaten thanks

PR #130, opened 2015-04-28
Assignment 4

zerok:


lib/index.js
```diff
@@ -2,11 +2,7 @@
 
 var Hapi = require('hapi');
 var Version = require('./version');
-
-
-// Declare internals
-
-var internals = {};
```
AdriVanHoudt: why did you remove internals?

lib/private.js
```diff
@@ -0,0 +1,59 @@
...
+
+var Basic = require('hapi-auth-basic');
+var Util = require('util');
+var Hoek = require('hoek');
+
+
+// Declare internals
+
+var internals = {};
+internals.users = require('./users.json');
```
AdriVanHoudt: this doesn't need to be in internals, better to just declare like hoek for example

AdriVanHoudt:
general linting issues
```js
Linting results:
	lib/private.js:
		Line 23: no-else-return - Unexpected 'else' after 'return'.
		Line 31: space-before-function-paren - Missing space before function parentheses.
	test/private.js:
		Line 14: hapi/hapi-scope-start - Missing blank line at beginning of function.
		Line 14: space-before-function-paren - Missing space before function parentheses.
		Line 89: space-before-function-paren - Missing space before function parentheses.
```

lib/index.js
zerok: Because they were unused and therefore linting complained :)

lib/private.js
```diff
@@ -0,0 +1,59 @@
...
+                handler: function (request, reply) {
+
+                    return reply(Util.format('Hello, %s!', Hoek.escapeHtml(request.auth.credentials.username)));
+                }
+            }
+        });
+
+        return next();
+    });
+
```
AdriVanHoudt: no need for newline here

lib/private.js
```diff
@@ -0,0 +1,59 @@
...
+        server.route({
+            method: 'GET',
+            path: '/private',
+            config: {
+                description: 'Returns a welcome message to the authenticated user',
+                auth: 'simple',
+                handler: function (request, reply) {
+
+                    return reply(Util.format('Hello, %s!', Hoek.escapeHtml(request.auth.credentials.username)));
```
AdriVanHoudt: is the username html? why escape it?

lib/start.js
```diff
@@ -4,14 +4,8 @@ var Hoek = require('hoek');
 var Server = require('./index');
 
 
-// Declare internals
-
-var internals = {};
```
AdriVanHoudt: again why remove this? This is here due to the hapi style guide

lib/private.js
zerok: Because a username in general should be considered user-provided data and therefore should be escaped.

AdriVanHoudt: Ok I can see that, imo a bit unnecessary here but ok

lib/index.js
AdriVanHoudt: are you using the build in lab linter?

zerok:
@AdriVanHoudt This is just plain weird. Seems like eslint is getting mixed into the linting result or something. I was running the linter constantly and got no compains in the end. Will fix them in the evening.

AdriVanHoudt:
@zerok do you have the latest version of Lab? There were some rules added recently

zerok:
@AdriVanHoudt I should have. The issue was that I also had a self-made .eslintrc file lying around which influenced what the linter was doing.

AdriVanHoudt:
Ah that was probably it. you can always see what the test results are on travis to see the results in a 'clean' environment

lib/private.js
```diff
@@ -0,0 +1,57 @@
...
+
+
+// Declare internals
+
+var internals = {};
+
+
+internals.validate = function (username, password, callback) {
+
+    var matchingUsers = Users.filter(function (account) {
```
hueniverse: This is far too complicated. Just do a key lookup by username and then check for the password value.

PR #138, opened 2015-05-25
Solved assigment 5 

roelof1967:


package.json
```diff
@@ -14,10 +14,12 @@
   ],
   "license": "BSD",
   "bugs": {
-    "url": "https://github.com/hueniverse/hueniversity/issues"
+    "url": "https://github.com/hapijs/university.git"
   },
-  "homepage": "https://github.com/hueniverse/hueniversity",
+  "homepage": "https://github.com/hapijs/university.git",
   "dependencies": {
+    "glue": "^2.0.0",
```
AdriVanHoudt: make sure to use the same notation here as the rest of the deps (2.x.x)

test/index.js
```diff
@@ -27,7 +31,7 @@ it('starts server and returns hapi server object', function (done) {
 
 it('starts server on provided port', function (done) {
 
-    Hueniversity.init(5000, function (err, server) {
+    University.init({connections: [{port: 5000}]}, {}, function (err, server) {
```
AdriVanHoudt: when using inline objects put spaces after { and before }, and maybe also useful to declare this in a separate variable to make it more readable

test/index.js
```diff
@@ -57,3 +61,26 @@ it('handles register plugin errors', { parallel: false }, function (done) {
...
+    }
+};
+
+internals.composeOptions = {
+    relativeTo: Path.resolve(__dirname, '../lib')
+};
+
+
+
+
```
AdriVanHoudt: random empty lines ;)

test/index.js
```diff
@@ -27,7 +34,10 @@ it('starts server and returns hapi server object', function (done) {
 
 it('starts server on provided port', function (done) {
 
-    Hueniversity.init(5000, function (err, server) {
+    var manifest = {connections: [{port: 5000}]};
```
AdriVanHoudt: I would go with default object notation like you did with internals.manifest below

roelof1967:
Chips. ,more then i wanted, This schould stop at commit 8626de07a407

PR #141, opened 2015-05-27
Assignment #5 - Glue and HTML views

### LINKED ISSUE #136: Glue and HTML views
#### We transitioned to hapijs university.

For details about this change see: <br/>
* New [README](https://github.com/hapijs/university) for explanation of changes.
* [This Discussion] (https://github.com/hapijs/university/issues/132). To see current thoughts about project direction.
* And this [Documentation](https://github.com/hapijs/university/issues/133)

Let's keep our project going! This is community learning which means much of the learning will be 
horizontal and not just vertical! Make your contribution, ask questions, give advice (review others code) , share your thoughts about our project's direction and let's push each other to keep growing :-) 

## The Assignment

*  Refactor the application to use the hapijs server composer  [hapijs / glue](https://github.com/hapijs/glue).  <br/>Glue should be used to load "version" and "private" plugins.
Here is a helpful resource from the Glue project: [Sample Manifest JSON object](https://github.com/hapijs/glue/blob/master/lib/index.js) . 

*  Create a plugin called "home":
   * This plugin should contain one endpoint called "home".
   * Configure the plugin to use handlebars template rendering engine to render html templates.
   * Configure this plugin to serve html view files and make the home endpoint render an html page/template named home.
   *  Configure the home plugin to find view files in a directory named "views" located in
   the root path of the application. For now, just make html pages for view files (no .hbs files).
 * The view directory should contain one html template named home.html. <br/>
    When viewed in a browser, the home.html template should display the path to this view file from the root of our project. 
*  In the end, the application should have thee endpoints. One called: "home", 
   the other called "version", and one called "private". Each point is located in it's own plugin 
   which glue composes when the server starts up. 
*  Ensure 100% test coverage.
*  Refactor project to meet new requirements https://github.com/hapijs/contrib/issues/31.  
   - Remove Makefile and switch to using npm scripts (see hapi package.json)
   - Remove root index.js (point package.json to lib/index.js)
   - Replace package.json licenses array with single license key and `"BSD-3-Clause"` value
   - Update code style to stop putting functions on single line (new lab will error on that lint warning)
* An extra challenge, after doing the above, refactor hapi-auth-basic registration code: 
   * Refactor the application to load hapi-auth-basic plugin using glue's manifest file.  
   * Remove hapi-auth-basic plugin registration code from the private.js file.
   * private.js plugins' functionality will not change.  We are just moving where the 
     hapi-auth-basic plugin is registered.  
   * Configure Glue to load hapi-auth-basic plugin and make a file dedicated to register the hapi-auth-basic strategy.
   Glue will reference this file in the manifest file to load the plugin. 
   * While making the assignment, there was discussion about why not to make this change. See: [The Discussion](https://github.com/hapijs/university/issues/132) for details. 
   But, I encourage everyone to try it.  Doing this exercise will help you get to know Glue and hapijs' auth plugin design better ;-)
   If you get stuck here, we will work together to make it work! 

#### What is coming next?
* adding static assets like css, images, js files.
* validation with joi
* sessions and authentication with hapi-auth-cookie
* See "What we are going to build?" section in the README https://github.com/hapijs/university  for the big picture plan.
* If you have suggestions for issues to study please make them!
Cheers!

hussainanjar:
Completed assignment 5.

Kept Glue manifest in external json file under config folder.

Closes #136


lib/home.js
```diff
@@ -0,0 +1,41 @@
...
+        relativeTo: __dirname,
+        path: '../views'
+    });
+
+    server.route({
+        method: 'GET',
+        path: '/home',
+        config: {
+            description: 'Home path',
+            handler: function (request, reply) {
```
AdriVanHoudt: why not use the view handler here?

lib/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "connections": [
+        {
+            "port": 8000,
+            "labels": ["web"]
```
AdriVanHoudt: not required

test/index.js
```diff
@@ -27,7 +31,7 @@ it('starts server and returns hapi server object', function (done) {
 
 it('starts server on provided port', function (done) {
 
-    Hueniversity.init(5000, function (err, server) {
+    University.init({connections: [{port: 5000}]}, {}, function (err, server) {
```
AdriVanHoudt: For readability it is better to extract the manifest variable and not notate it on 1 line imo

AdriVanHoudt:
I like the idea of putting the server config in a separate folder, will update my PR :P, but I wouldn't use it in tests, mainly for port use, this can change once we add Confidence and have environment specific configs

AdriVanHoudt:
Also something that I did was to have a default server and then instead of just passing full on manifests, only pass along what you want to change to the default on and then do Hoek.applyToDefaults

lib/manifest.json
```diff
@@ -0,0 +1,13 @@
...
+        }
+    ],
+    "plugins": {
+        "./version": null,
```
AdriVanHoudt: since the value of the plugin is the options you want to pass it more logical to pass along {} but this will work to

lib/home.js
hussainanjar: No good reason, just like that. Though I agree view handler is the better option :smile:

lib/manifest.json
hussainanjar: yep, was playing with it forgot to remove.

test/index.js
hussainanjar: Makes sense

lib/manifest.json
hussainanjar: I always prefer null over empty objects, I feel they are more error prone.

AdriVanHoudt: true but in this case you want to pass no options, so an options object with nothing in it, just personal preference but both are fine

lib/home.js
AdriVanHoudt: also typo s/optinos/options

hussainanjar:
@AdriVanHoudt I agree with the idea of having manifest.json in config folder but I'm not sure about the idea of defaultServer.

One config file defining everything looks more neat and will be easily extendible if different connections are required for web and say API.

More readable too I think.

AdriVanHoudt:
@hussainanjar those are the same thing right? Only I allow to override stuff and you don't

hussainanjar:
@AdriVanHoudt why do you feel the need of defining default server outside of config file ?

hussainanjar:
@AdriVanHoudt got it now while writing tests, your approach makes test simpler. :+1:

Issue #144, opened 2015-06-01
Static Assets, Partials, and TLS


zoe-1:
#### Configure the application to use TLS/SSL
- Generate your own TLS/SSL certificate                                                                    
- Make "certs" directory in ./lib.                                                               
- Place the certificates you generate into "certs" directory.          

#### Configure two server connections
 - One labelled ['web'] on port 8000 with no TLS configured.                                      
 - The other labelled ['web-tls'] on port 8001 will use TLS.                                      
      Configure this connection to use the ssl certificates you just generated                       
      and placed in ./lib/certs                                                                      
    * Example server certificate configuration at [npm newww ](https://github.com/npm/newww/blob/13e26e0eaf5c6699f068bb5418d932af76be4aef/server.js)

#### Use TLS everywhere
Configure the application to always use TLS [Transport Layer Security] (http://en.wikipedia.org/wiki/Transport_Layer_Security).  <br/> Every http request should be redirected to https. 
Two references: 
* [@hueniverse on the issue](https://github.com/hapijs/hapi/issues/778)
* [related StackOverflow post](http://stackoverflow.com/questions/28650829/hapijs-using-both-http-and-https-on-one-connection)

### Configure the home plugin.
  - Make a new point in the home plugin called: "login"
    * Make login route display view template named login.html.
  - Use partials to display the login form
    * make a directory inside of the ./views called partials                          
    * configure the home plugin to use the ./views/partials directory to load partials from. 
        * In ./views/partials/ directory make a file named form.html.
           * form should have two inputs one type="text" named "username" and the second type="password" named "password".
          * form should have one submission button with text value "Login".                        
        * Inside login.html template load the partial file named "form.html" 

#### Use Static Assets                                                                          
- Make directory in root of project called "assets".                                            
- Inside the assets directory, make three directories:                                           
  "styles" for css, 
  "scripts" for JS scripts,                                                                      
   "images" for project images.                                                                   
    * Place Hapi University logo into the images directory                                           
      and display it on the home page (home.html). 
      (image is in the repo ./images/logo.png)
    * Place a CSS file in the stylesheets directory (styles).                                        
      Name css file styles.css:
       * Make one simple style:                                                                      
         h3 { font-size: 24px; font-weight: bold; }                                                  
       * Use the stylesheet in home.html. Make a title inside h3 tags.
    * Make a script named login.js and put in ./assets/scripts   
      * Make this script print a console message to the browser when the form button is clicked.                                    
[...]

PR #151, opened 2015-06-08
- Completed assignment "Static Assets, Partials, and TLS" issue #144

luismatute:


luismatute:
Done

luismatute:
Weird, npm test threw that everything was alright... I'm new to testing, sorry for the mistakes. Hopefully it works as expected now :)

screen shot 2015-06-15 at 8 54 14 am

zoe-1:
@luismatute Probably, need to upgrade lab.  
Then, this will be fixed. Sometimes lab gets upgraded and they make it more strict 
then, older versions pass while newer versions fail.  It's happened to me before. 
I just upgraded so had listing errors.

AdriVanHoudt:
Mainly my bad haha sowry :D if I have time there will be an update this week even `break all the things!` :D

PR #152, opened 2015-06-09
Assignment6

rutaihwa:
Submitting assignment6 #144


zoe-1:
@rutaihwa This is nice work.
The only thing I would change would be to take more defensive measures on testing tls routes as I 
mentioned in the my post critiquing assignment6 https://github.com/hapijs/university/issues/144.

rutaihwa:
@zoe-1 Yeah, I would beef up the tests.

lib/start.js
```diff
@@ -10,13 +11,27 @@ var internals = {};
...
+            labels: ['web-tls'],
+            tls: Config.tls
         }
     ],
     plugins: {
-        './version': {},
-        './private': {},
-        './home': {},
+        './version': [{
+            'select': ['web', 'web-tls']
```
AdriVanHoudt: why register the plugins also on web when all routes are redirected to web-tls?

test/home.js
```diff
@@ -19,17 +20,33 @@ var it = lab.test;
...
+
+                expect(res.statusCode, 'Status code').to.equal(301);
+                expect(res.headers.location).to.equal('https://localhost:8001/home');
+
+                server.stop(done);
+            });
+        });
+    });
+
+    it('returns an home page via https', function (done) {
```
AdriVanHoudt: where is the actual test for testing if it indeed is served via https?

rutaihwa: @AdriVanHoudt What you mean to this test or the above one?

lib/start.js
rutaihwa: @AdriVanHoudt, I think you are right :+1: , since everything gets caught and redirected to `https` before being routed, it may be not necessary to register `web`.

test/home.js
zoe-1: @AdriVanHoudt  expect(res.headers.location).to.equal('https://localhost:8001/home');
Headers location shows it was redirected to https, proving it was redirected to https.

zoe-1:
@AdriVanHoudt  Good to have you back making comments again. :-)
Have not seen you in a couple days.

lib/start.js
AdriVanHoudt: seems unnecessary at least

test/home.js
AdriVanHoudt: below, i see you did in other tests although if you change the port your test fails even when routing to https which seems weird but also will probably never happen ^^

rutaihwa: @AdriVanHoudt How would you test it?

lib/start.js
rutaihwa: Agree, will be addressed in the next commit.

test/home.js
AdriVanHoudt: @rutaihwa according to http://stackoverflow.com/a/10353248/2339622 you should be able to get it via `res.result.raw.req.connection.encrypted`

AdriVanHoudt: @rutaihwa turns out it's not really available, because of the redirect the protocol in the connection info is http (somewhere deep in the res object ;)) so the better thing would be to do `res.headers.location.substring(0,5) === 'https'` I think

PR #157, opened 2015-06-17
Assignment1 

bsocius:
Created a server which response with a json object and gives the version of the current project.


zoe-1:
Congratulations on your first PR !

AdriVanHoudt:
:tada: 
2 things: I personally prefer the 8.x.x notation for versions and what is that `package.json~` thing?

bsocius:
What is the difference between my notation and your notation, it seems that both install the compatible version of the package. But i am not completely sure. The package.json~ file or file.extension~ is something that my editor is producing, Its a backup file, So whenever i do i change and save it to disk, the old file is saved as file.extension~.

I have to put those files into the .gitignore file so it will not pollute my git repo.

AdriVanHoudt:
it does the same only personally i find the x notation more clear than vague symbols like ^ and ~

Issue #155, opened 2015-06-16
Validation, Authentication, and Authorization


zoe-1:
## Objectives
* Use joi to validate submitted POST data.
* Gain basic understanding of the hapijs request lifecyle.
* Understand hapi-auth-cookie authentication and authorization
    * strategies
    * sessions and cookies
    * scopes
* Implement simple RBAC (role based access controls) using hapi's scopes
  to implement an authorization scheme.
* Make basic login and logout functionality.
* Configure application to handle bad route requests.
* Begin designing an api to support multiple clients.

### Study Helps
If you read the assignment requirements and do not know where to start then go to: [the study helps](https://github.com/hapijs/university/issues/153).

###  Cookie-based session management.
Configure hapi-auth-cookie for cookie-based session management.
Every route will be black listed requiring authentication and authorization globally.
White list public routes. Make two user roles for basic RBAC authorization:
* "admin" the super user
* and "user"  basic user

There will be three types of authorization restrictions:
  * "public" routes open to all
  * "user" routes restricted to authenticated users with the role of "user" or "admin".
  *  "admin" routes restricted to authenticated users with the role "admin".

### Login Process
We created the GET /login route in assignment6. This route is to be open to the public.
When the user clicks the submit button on the HTML form, client side JavaScript should handle the submission and make an AJAX POST request. If the POST request is successful, the login form should disappear and user is presented with links to access restricted content /account or /logout.  At this stage the login process is complete, session cookies are set, and authorized users can begin to access restricted content.

If the login AJAX request fails, a red HTML format error message is displayed above the login form.
If the user clicks submit again the error message should disapear.
Bonus: Add clientside logic to avoid duplicate form submissions.

#### api first steps
Building towards a client-server architecture for an application that serves multiple clients.
Begin building the api:
* Make an 'api' label along with the web-tls connection.
* Make a ./lib/api directory.
* Make a ./lib/api/login.js
    * Make POST ./login route in this directory.
    * configure this plugin to use hapi-auth-cookie and begin the session upon successful login.
    * handle AJAX POST request at ./login route.

#### The /account route
After a successful login, user sees "My Account" link and "Logout" link.
The "My Account" link goes to the /account route which is restricted to authenticated users with roles of "user" or "admin".  The /account route will display a a page with the username of the logged in user and the following links: "Logout", "Home", and "Admin Panel"  (if user has admin privileges).
Correspondingly, the previous mentioned links point to /logout, /home, and /admin.

#### The /admin route
[...]

PR #158, opened 2015-06-26
Assignment7

zoe-1:
Meets requirements defined in assignment7 #155. 
However, would like to improve:
  * Add before after logic to tests to make them cleaner.
    and spend time making them prettier.  
  * Browser caching becomes a pain at times.
  * Better use of internals and global abatement.
  * Using too many global variables
    in a program weakens it's resiliency of programs and should be avoided.
    This is what hapijs is doing with internals inside modules.
    All module global variables are placed within internals. This
    controls how many variables are global (global abatement).
    Douglas Crockford - JavaScript: The Good Parts p.25  
  * For example, client side code could be improved to utilize hapi's style of global abatement.
  * Use @FennNaten 's sytle of server.dependency([dependencies], * after) logic instead
     of my current style of calling the callback.  His is more elegant.
  * how handlebars is being used, currently we just add a  partial
      but there are many handlebar features that could be added.


zoe-1:
When building this project did not test v0.10.x of node.js.  After making the PR, travis failed because 
v0.10.x cannot use JavaScript Promises.  So, made a quick fix and remove Promise logic from the project. This is unfortunate because the Promise logic modeled after @poeticninja was nice and I made JavaScript promises part of the assignment.

mindbits:
can't you just use bluebird or another promise library?

lib/home.js
```diff
@@ -22,53 +29,126 @@ exports.register = function (server, options, next) {
...
+                // Admin check
+                var administrator = null;
+
+                if (request.auth.credentials.scope[0] === 'admin') {
+                    administrator = true;
+                }
 
-    ]);
+                var userInfo = { first: request.auth.credentials.first, last: request.auth.credentials.first };
+
```
rutaihwa: shouldn't it be `, last: request.auth.credentials.last` ? typo

PR #160, opened 2015-07-06
Assignment1

benleen:


lib/index.js
```diff
@@ -0,0 +1,17 @@
+var Hapi = require('hapi');
+var pjson = require('./../package.json');
```
AdriVanHoudt: hapi guidelines state that all imported modules should be capitalized
as a personal note I would use more descriptive names as `Package` or `Pkg` if you really want it short, `pjson` doesn't really tell me what the object is

zoe-1:
@benleen Travis fails because you are getting below linting errors:
* hapi/no-shadow-relaxed - server is already declared in the upper scope.
* hapi/no-shadow-relaxed - next is already declared in the upper scope.

#### Make .eslintrc file in project root to fix shadow linting issues
.eslintrc file content below:
``` JavaScript
{
    "extends": "./node_modules/lab/lib/linters/eslint/.eslintrc",
    "rules": {
        // Override any settings from the "parent" configuration
        "hapi/no-shadow-relaxed": [1, { "ignore": ["err", "done", "server", "next", "res"] }],
    }
}
```
Above will make Travis hapi :-)

zoe-1:
You are adding requirements from later assignments into assignment1.  
For example, you are using Glue in assignment1 but that was not added until 
assignment5.  This is not a big deal as long as you are aware of it.

zoe-1:
See comments [here](https://github.com/hapijs/university/issues/155) about configuring 
linting options for more details about shadow linting.

benleen:
@zoe-1 
* Should I care about Travis for assignment1? If I'm not mistaking it's even explicitly mentioned for assignment3. I can live with the red cross. ;)
* About the branches. That seems to be what I've done... I must be missing something.
Branch in my fork of hapijs/university: https://github.com/benleen/university/tree/assignment1 from which my PR originates. Also... I don't seem to using "Glue" either.

zoe-1:
@benleen  I looked over your assignment a bit too quickly.  
Yes, you are right about the branches.  I was looking at your master branch. Sorry. 
In respect to Travis, I would recommend getting travis to pass. But, since the original assignment
did not require it, no worries if you do not get it to pass.

PR #166, opened 2015-07-10
Assignment3

benleen:


test/index.js
```diff
@@ -0,0 +1,42 @@
...
+            Version.register = backup;
+
+            return next(new Error('plugin failed to register'));
+        };
+
+        Version.register.attributes = {
+            name: 'fake version'
+        };
+
+        Server.init(5000, function (err, server){
```
AdriVanHoudt: do not set the port if not needed for the test this will only allow for errors

zoe-1:
No responses to comments and discussion so closed this.

PR #170, opened 2015-07-26
[Assignment 2] Convert to a plugin

johnmanko:


johnmanko:
Not sure why it failed.  Here is the message:
rake aborted!
No Rakefile found (looking for: rakefile, Rakefile, rakefile.rb, Rakefile.rb)
/home/travis/.rvm/gems/ruby-1.9.3-p551/bin/ruby_executable_hooks:15:in `eval'
/home/travis/.rvm/gems/ruby-1.9.3-p551/bin/ruby_executable_hooks:15:in `<main>'
(See full trace by running task with --trace)

AdriVanHoudt:
`Could not find .travis.yml, using standard configuration.` Which is ruby I think and not Node so it will try to do the wrong thing :P

zoe-1:
Nice work.  Keep up the good work.

PR #182, opened 2016-02-04
[Assignment 4] Basic Authentication

### LINKED ISSUE #118: Basic authentication

apoorvakorde:
closes #118


zoe-1:
nice work. :-)

PR #186, opened 2016-02-14
[Assignment 4] Multiple callbacks or thrown errors received

jd73:
I have an implementation of Assignment 4 that passes all the individual test cases that I have for it but gives the following error as a test script error.

> Multiple callbacks or thrown errors received in test "/private returns an error if it fails to register auth" (error): Expected undefined to exist
>       at /Users/joe/projects/university/test/private.js:92:28
>       at /Users/joe/projects/university/lib/index.js:22:94
>       at Server.g (events.js:273:16)
>       at emitNone (events.js:85:20)
>       at Server.emit (events.js:179:7)
> 
> There were 1 test script error(s).

Other than the test suite not finishing successfully it satisfies all the of the requirements of the assignment as far as I can tell.

I've spent a fair amount of time trying to figure out where the extra error or callback may be coming from. I thought that I might have missed a 'return' in front of a callback call somewhere or something but I haven't been able to find anything like that. I also tried running the tests with the '-d' flag to try to debug where it may have been coming from but all my test cases failed when I did that.

I'm stumped at this point. Any suggestions for what I might be able to do to figure this out?


lib/private.js
```diff
@@ -0,0 +1,55 @@
...
+                handler: function (request, reply) {
+
+                    const html = '<div>Hi ' + request.auth.credentials.username + '!</div>';
+                    return reply(html);
+                }
+            }
+        });
+    });
+
+    return next();
```
apoorvakorde: this return is at the wrong place. it should be at the end of the callback to the server.register used above

jd73: Thank you!

PR #202, opened 2016-03-11
Assignment8 cleanup.

zoe-1:
* Added documentation updates from master after review.
* changed /lib/cookie-auth to ./lib/auth-cookie.js


zoe-1:
Did not realize all these commit were going to be merged. 
Canceling this merge will make a cleaner one.

PR #211, opened 2016-03-26
My assignment1

kernkonzentrat:


package.json
```diff
@@ -0,0 +1,31 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/university.git"
+  },
+  "keywords": [
+    "hapijs"
+  ],
+  "author": "Robin Rush",
+  "license": "BSD",
```
zoe-1: hapi always uses:
`BSD-3-Clause`

package.json
```diff
@@ -0,0 +1,31 @@
...
+  "bugs": {
+    "url": "https://github.com/kernkonzentrat/university/issues"
+  },
+  "homepage": "https://github.com/kernkonzentrat/university#readme",
+  "dependencies": {
+    "hapi": "^13.2.2"
+  },
+  "devDependencies": {
+    "code": "^2.1.1",
+    "eslint-plugin-markdown": "^1.0.0-beta.1",
```
zoe-1: eslint-plugin-markdown is not part of university project. 
lab and code use eslint configured to the hapi style guide.

.eslintrc
```diff
@@ -0,0 +1,7 @@
+{
```
zoe-1: we do not extend hapi linting in the university project.

lib/index.js
```diff
@@ -0,0 +1,35 @@
...
+});
+
+internals.server.start((err) => {
+
+    Hoek.assert(!err, err);
+
+    console.log('Server running at: ', internals.server.info.uri);
+});
+
+module.exports = internals.server;
```
zoe-1: no need to export the hapi server object.  
There are no other applications or modules that are going to consume the server, so the export is irrelevant.

package.json
```diff
@@ -0,0 +1,32 @@
...
+    "url": "https://github.com/kernkonzentrat/university/issues"
+  },
+  "homepage": "https://github.com/kernkonzentrat/university#readme",
+  "dependencies": {
+    "hapi": "^13.2.2",
+    "hoek": "^3.0.4"
+  },
+  "devDependencies": {
+    "code": "^2.1.1",
+    "eslint-plugin-markdown": "^1.0.0-beta.1",
```
zoe-1: eslint-plugin-markdown is not part of project.

test/server.spec.js
```diff
@@ -0,0 +1,28 @@
...
+lab.experiment('hapijs_uni', () => {
+
+    lab.test('Get /version should return 200 and current project version', (done) => {
+
+        const options = {
+            method: 'GET',
+            url: '/version'
+        };
+
+        Server.inject(options, (response) => {
```
zoe-1: The way you built your server in ./lib/index.js is not conducive for 
building tests. Study the following two scripts for understanding how 
to build your server and get test coverage:

https://github.com/hapijs/university/blob/assignment3/test/version.js
https://github.com/hapijs/university/blob/assignment3/lib/index.js

Notice the relationship between:
./lib/index.js
```
exports.init = function (port, next) {

    const server = new Hapi.Server();
    server.connection({ port: port });
    server.register(Version, (err) => {

        if (err) {
            return next(err);
        }

        server.start((err) => {

            return next(err, server);
        });
    });
};
```
and ./test/version.js
```
University.init(0, (err, server) => {
 .....
}
```

The init method in ./lib/index.js returns the error and server object
of the initiated server which is needed to test to the server. 
```
(err, server) => {
 ......
}
```

Note: this assignment does not require tests but you tried here so made some comments to help.

lib/index.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+const Hapi = require('hapi');
+const Hoek = require('hoek');
+
+const internals = {};
+
+internals.package = require('../package.json');
+internals.response = { version: internals.package.version };
+internals.server = new Hapi.Server();
```
zoe-1: The university style is to create a variable just for the server.
The application we are building is a server and internals will be consumed inside the server. So, the preference is to do as follows:
```
const server = new Hapi.Server();
```

PR #215, opened 2016-04-07
Assignment1 completed and works fine 

despirados:


zoe-1:
This assignment fails. There is not start command or script.

lib/version.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+'use strict';
+
+const Package = require('../package.json');
+
+
+// Declare internals
+
+const internals = {
+    pluginName: "Version",
```
zoe-1: do not need to name your plugin in internals.

PR #224, opened 2016-10-16
Refine and clarify, then share :)

eriktrom:
Refine and clarify code having to do with use of glue, testing and logging

This PR contains more than one concern. It is also based off of assignment 8. It may therefore not be something to merge as is. If there is interest in one or more parts, I don't mind picking this apart according to feedback. Fyi.

Loose Notes:

- Upon initially learning Hapi earlier this year there we're some
  things that led me to eventually remove Glue from my app. This later made
  testing difficult. Coming back to this repo led me to see what confused
  me - which mainly revolved around pluginOptions vs registerOptions. Glue
  uses the registerOptions while pluginOptions can be passed to a plugin 
  from the main app. This is especially useful when working with connectionless
  plugins.

- Along the way I also saw some patterns dealing with the log events + use
  of good that helped me look at my own app logging situation a bit differently.

- I also noted the use of small things like using process.cwd() over `__dirname`
  that I found useful in my own server when working with babel/typescript +
  rollup. 

As I saw such things I made an attempt to clarify them. The changes are all
benign meaning there is no new functionality, just code re-organization.

Feel free to use what seems useful or the whole branch in further
assignments/refactors.

Thanks for the awesome example btw. Without it I would never starting 
using Hapi all those months ago.

### DIFF
--- a/lib/lout.js
+++ b/lib/lout.js
@@ -14,7 +14,8 @@ exports.register = (server, options, next) => {
     // Earlier versions of hapi included vision and inert out of hapi. Now,
     // these must be included seperately.
 
-    server.register({ register: Lout, options }, err => {
+    server.register({ register: Lout, options }, (err) => {
+
         if (err) {
             return next(err);
         }
--- a/lib/lout.js
+++ b/lib/lout.js
@@ -14,18 +14,9 @@ exports.register = (server, options, next) => {
     // Earlier versions of hapi included vision and inert out of hapi. Now,
     // these must be included seperately.
 
-    server.dependency(['inert', 'vision'], internals.after(server, next));
-
-    // return next();
-};
-
-
-internals.after = (server, next) => {
-
-    server.register(Lout, (err) => {
-
-        if (!err){
-            return next();
+    server.register({ register: Lout, options }, err => {
+        if (err) {
+            return next(err);
         }
 
         return next();
@@ -33,5 +24,8 @@ internals.after = (server, next) => {
 };
 
 exports.register.attributes = {
-    name: 'Lout'
+    name: 'Lout',
+    dependencies: ['inert', 'vision']
 };
+
+
--- a/lib/start.js
+++ b/lib/start.js
@@ -11,6 +11,12 @@ const Config = require('./config');
 
 const internals = {};
 
+internals.goodFilePath = (() => {
+    const env = process.env.NODE_ENV;
+    const name = env ? 'good '+env+'.log' : 'good.log';
+    return './log/'+name;
+})();
+
 internals.manifest = {
     connections: [
         {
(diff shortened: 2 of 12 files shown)


eriktrom:
PS - the pattern for cleaning up the log file after tests runs seems like it could be cleaner. Open to feedback.

zoe-1:
Really nice work :-)
Looked at your PR some today and appreciate what you have done.  
Over the next couple days will provide more feedback and interaction. 
Cheers!

eriktrom:
@zoe-1 no rush - i'll be OOO till the 1st so take ur time

zoe-1:
@eriktrom I have been looking over your work. In general, I like all improvements you made. Very nice work!  However, in respect to `__dirname` versus `process.cwd()` I have a few thoughts.  

* Does your `server babel/typescript + rollup` have issues with __dirname?  

* I prefer __dirname when building services. Using __dirname allows to restart services from a BASH script not located in the root of the app.  __dirname always points to the directory of the file it was written in. On the other hand, process.cwd() points to the directory of the process executing the script. This causes issues when the app is started with a script outside the root of the application and breaks the paths.  

Any thoughts?  

Lately, in my use of hapi I have been trying not to write big monolithic apps but instead writing a bunch of little apps / processes that talk to each other.  Using process.cwd() would not work well for me because it would stop my ability to use a script to restart multiple services at the same time. 

In respect to writing multiple little apps that talk to each other, I found this [video](https://www.youtube.com/watch?v=A7ib9ZGx1Rs) by @substack interesting. Perhaps, you will like it.  I think hapi and nodejs really shine when building many small apps / services that talk to each other.  And, hapi's [h2o2](https://github.com/hapijs/h2o2) is really powerful for building proxies that consume multiple services.  npmjs' CTO gives a good talk on nodejs and SOA service oriented architecture [here](https://www.youtube.com/watch?v=5yK3lx-PQV0). note: npmjs uses hapijs. 

Anyhow, I don't think using process.cwd() or __dirname is really that important of an issue. Every project has it's own unique needs.  Depending on the context one or the other would work fine.  Just trying to share some meaningful feedback after you shared your meaningful work.

lib/lout.js
```diff
@@ -14,24 +14,18 @@ exports.register = (server, options, next) => {
...
         }
 
         return next();
     });
 };
 
 exports.register.attributes = {
-    name: 'Lout'
+    name: 'Lout',
+    dependencies: ['inert', 'vision']
```
zoe-1: Declaring decencies in attributes does not guarantee dependencies will be loaded first. 
See http://hapijs.com/api#serverdependencydependencies-after for details.
My preference would be to use dependencies after to bullet proof dependencies being loaded first.

lib/start.js
```diff
@@ -60,8 +66,14 @@ internals.manifest = {
...
-            plugin: './crumbit',
-            options: {
+            plugin: {
+                register: './crumbit',
+                options: { // pluginOptions - passed as 'options' to plugin
+                    restful: true,
+                    cookieOptions: { isSecure: true }
+                },
+            },
+            options: { // registerOptions - NOT passed to plugin; instead passed to Glue under the hood
```
zoe-1: Thank you for making the comments here in the manifest and explaining this. 
Definitely improved the university :-)

PR #226, opened 2016-10-27
@eriktrom improvements & some modifications

zoe-1:
#### merging @eriktrom 's contributions into master:
* better use of glue's manifest options. 
Clarifies difference between glue manifest options versus plugin options.
* server logging improvements. 
* see #224 for more details.


eriktrom:
Looks good @zoe-1. 👌

PR #254, opened 2017-08-14
Assignment9

zoe-1:
* request lifecycle and route level extensions
* summary of request lifecycle (big picture)


devinivy:
This is nice!  I'd like to dive a little deeper, but as a first pass all I can say is it seems to have a whole bunch of accurate information regarding extensions :)  I like that you've referenced sources from all over the hapi-verse.

zoe-1:
@devinivy Thank you for looking the assignment over :)

PR #264, opened 2018-07-02
My assignment1

LeonAppDev:
# Handler changed in hapi version 17.5.2
Since the right answer still use old version 16's method to create server and reply, need a new version to refelect changes on hapi 17.5.2

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,0 +1,26 @@
+const Hapi = require("hapi");
+
+(async ()=>{
+
+       const server = Hapi.server(
+         {
+           host:"localhost",
+           port:Number(process.argv[2]||"8000")
+         }
+       );
+
+      server.route({path:"/version",method:"Get",handler:(request,h)=>{
+
+          return "version";
+
+      }
+
+      }
+      );
+
+   await server.start();
+
+   console.log("Hapi server starts: ",server.info.port);
+
+
+})()
--- a/lib/index.js
+++ b/lib/index.js
@@ -4,31 +4,35 @@
 
 const Hapi = require('hapi');
 const Package = require('../package.json');
+
+
 // Make declaration
-let internals = {
-  version: {version: Package.version},
+const internals = {
+    version: {version: Package.version},
 };
 
 internals.start =async ()=>{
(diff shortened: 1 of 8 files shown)


lib/index.js
```diff
@@ -6,23 +6,22 @@ const Hapi = require("hapi");
 const Package = require("../package.json");
 //Make declaration
 
-(function (){
+(async ()=>{
```
zoe-1: hapi style guide does not use [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) function calls to start the server.
See: [style guide regarding module globals](https://github.com/hapijs/contrib/blob/master/Style.md#module-globals)

lib/index.js
```diff
@@ -6,23 +6,22 @@ const Hapi = require("hapi");
...
 
-       const server = new Hapi.server();
+       const server = new Hapi.server({port:Number(process.env.PORT||"8000")});
```
zoe-1: See hapi style rules [for spaces](https://github.com/hapijs/contrib/blob/master/Style.md#spaces)) regarding spaces in objects `{}`.

zoe-1:
See [sample solution](https://github.com/zoe-1/university-dev/blob/v0.1.1/lib/index.js) for more details.
In respect to how the hapi server is built and started, we need to build the server as a function which can be exported. Therefore, using the [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) function call does not work.  Future assignments will export the 'init' or 'start' function to a start script and test scripts.  Plus, if we plan to build multiple hapi services which communicate with each other, it is important to build the hapi servers / services in a way that another application can start and stop them (ie. manage the services). Exporting the init() allows for this.  This assignment does not require an init() to be exported. Just trying to explain some of the reasons for hapi's style preferences and why we don't use IIFE functions to start the server. Above [Module Global Styles](https://github.com/hapijs/contrib/blob/master/Style.md#module-globals) shows preferred style to init() the server.

LeonAppDev:
Hi, Zoe
    Thanks for your good explanation, I have another project to run so it takes me so long to code according to the convention. 
     I add ESLint and apply convention rules using it.

